PR #17, opened 2014-10-06
add a way to monitor multiple flume processes

mzupan:
This will add a way to monitor multiple flume processes per host. It also keeps the normal functionality of a single flume processes on a host. 

We run both setups where we have a single process and also 4 and this change passes the tests and also works in production.


kormoc:
Thanks for writing this, but the 4.0 branch supports running the same collector multiple times, so there's no need to build in support per collector anymore.

PR #20, opened 2014-11-04
ELBCollector: Thread collection by Zone

bmhatfield:
Threads the accesses done by the ELBCollector out at the zone level (after playing around with it, it's about as performant as at the ELB level, but uses far fewer threads and I hope this means that its less likely to get rate-limited by AWS in the future)

### DIFF
--- a/src/collectors/elb/elb.py
+++ b/src/collectors/elb/elb.py
@@ -44,6 +44,7 @@ import datetime
 import functools
 import re
 import time
+import threading
 from collections import namedtuple
 from string import Template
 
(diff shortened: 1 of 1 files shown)


kormoc:
So the 4.0/master branch supports running a collector multiple times, each independently. I'm gonna close this, but thanks for working on it!

bmhatfield:
I realize this is a bit of an old PR, but I'd like to reopen this with the suggestion that we merge it even though there also exists the new threading model in 4.0.

Specifically, the collector as written today is automatic for all ELBs with one configuration. Creating configuration files for each region won't really solve my problem (many ELBs in a single region), and being forced to list out each ELB where they would be otherwise automatically detected is worse.

This collector can continue to work totally automatically with this change and one nearly-stock configuration. I think the threading is easy to understand and get right in this context.

bmhatfield:
upload

For context, here's the improvement in collector time automatically gained by this PR.

@mattrobenolt @kormoc @analogue - Ya'll had weighed in before. I'm thinking of merging this because it moves the collector from "non-functioning in certain environments" to "works well". Based upon my explanation above of the assumption that collector threading does not actually help here, do you object?

mattrobenolt:
It seems reasonable to me. I don't have any strong objections, and the code seems fine without running it myself.

mattrobenolt:
I triggered a rebuild and tests look good.

bmhatfield:
Thanks for triggering the re-run!

I plan on merging this tomorrow morning (Eastern), so please do weigh in before then if there's something I have not understood @analogue @kormoc !

Thanks,
Brian

analogue:
Apoligies for chiming in late..

I think the PR is fine but think the default behavior should not be to use a thread per zone because of the worst case scenario.  Given AWS has nine regions and conservatively using two zones per region, that is  18 threads (I know, a bit of a stretch for someone to be in all available regions).  Nevertheless, it would be great if this was toggled by a config option instead and defaulted to not using a thread per zone.  Just my 2c :-)

PR #27, opened 2014-12-03
Add regex bean matching, and regex key search/replace to normalize names

sbrynen:
This patch adds two more config items to the JolokiaCollector.conf
mbeansre:  works like mbeans: but matches with regularexpressions
rewrite: which allows rewrite pairs in order to rename collected keys before data is passed to the handler


kormoc:
Can you also rebase your PR with current master? Thanks!

sbrynen:
it's all re-based now (and pep8 fixed)

sbrynen:
Any chance we can get this merged in now??

sbrynen:
I've created a new pull request as this was so old

PR #35, opened 2014-12-17
HttpJSON: Allow multiple host/query for the collector

eMerzh:
As there is no available way to execute multiple instance of a collector,

here is my take to update the httpJson collector to be able to load multiple url ...


kormoc:
The 4.0 branch supports running the same collector multiple times. Thanks for building this functionality, but I would rather simplify the collectors and run each one multiple times.

eMerzh:
No problems! 
I also prefer the more generic method ... hope 4.0 will be out soon ;-)

eMerzh:
@kormoc is there some documentation about how to run the same collector multiple time ?

PR #36, opened 2014-12-17
HttpJson collector: Allow usage of header configuration

eMerzh:
I wanted to be able to specify some headers on my httpjson collector,
as in the http one ...here is my small patch about it ..

It's useful for authorisation, for Host check, and plenty other things

### DIFF
--- a/src/collectors/httpjson/httpjson.py
+++ b/src/collectors/httpjson/httpjson.py
@@ -20,7 +20,8 @@ class HTTPJSONCollector(diamond.collector.Collector):
         config_help = super(HTTPJSONCollector, self).get_default_config_help()
         config_help.update({
             'url': 'Full URL',
-            'headers': 'Header variable if needed. Will be added to every request',
+            'headers': 'Header variable if needed. '
+                'Will be added to every request',
         })
         return config_help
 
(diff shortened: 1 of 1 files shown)


src/collectors/httpjson/httpjson.py
```diff
@@ -27,7 +28,8 @@ def get_default_config(self):
         default_config = super(HTTPJSONCollector, self).get_default_config()
         default_config.update({
             'path': 'httpjson',
-            'url': 'http://localhost/stat'
+            'url': 'http://localhost/stat',
+            'headers': {'User-Agent': 'Diamond HTTP collector'},
```
kormoc: Does this work as expected for setting in config files?

eMerzh:
now, it should be better ... i left a small typo ... 
but it seems to work as expected ...no?

eMerzh:
Any news on that?

bmhatfield:
Hi - This PR appears to have PEP8 issues. Would you mind cleaning that up?

eMerzh:
@bmhatfield you're right :) here it is ... :+1:

src/collectors/httpjson/httpjson.py
bmhatfield: I'm pretty happy with this change in general / prepping to merge this, but I'd like to second this question. Can you provide an example configuration snippet to confirm what it looks like?

eMerzh: If you set 

```

[[HTTPJSONCollector]]
enabled = True
measure_collector_time = False
url = http://localhost/HelloWorld
[[[headers]]]
X-Hello=World 
```
it should work as expected (set a header key: X-Hello and value World)

bmhatfield:
Thanks for this!

PR #39, opened 2014-12-24
Adding Netapp cluster functionality

deepakhj:
Cluster uses a seperate set of API's than vfiler mode. Seperated both calls into two functions. Set cluster=True to turn on cluster mode. I also updated the config/init functions to match other plugins.

### DIFF
--- a/src/collectors/netapp/netapp.py
+++ b/src/collectors/netapp/netapp.py
@@ -40,6 +40,7 @@ import unicodedata
 
 from diamond.metric import Metric
 import diamond.convertor
+from diamond.collector import str_to_bool
 
 
 class NetAppCollector(diamond.collector.Collector):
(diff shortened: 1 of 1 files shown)


src/collectors/netapp/netapp.py
```diff
@@ -190,16 +190,38 @@ class NetAppCollector(diamond.collector.Collector):
...
-        default_config['enabled'] = "false"
-        default_config['path_prefix'] = "netapp"
-        default_config['netappsdkpath'] = "/opt/netapp/lib/python/NetApp"
-        return default_config
+        """
+        Returns the default collector settings
+        """
+        config = super(NetAppCollector, self).get_default_config()
+        config.update({
+            'enabled': 'false',
```
kormoc: Shouldn't set enabled to anything anymore

kormoc:
Can you also update your PR with the current code? Thanks!

bmhatfield:
Hi - This PR appears to have PEP8 issues. Would you mind cleaning that up? Check the Travis build log for more information.

deepakhj:
Fixed pep8 and rebased.

bmhatfield:
@deepakhj Sorry for the back-and-forth here. I'm looking towards merging this, but I feel a little uncomfortable about the potential for backwards-incompatible changes here. For someone who's already using this collector in non-cluster mode, what kind of changes will they have to make? None?

deepakhj:
I kept the non cluster logic in it's current function and created a new function to handle the cluster metrics. You shouldn't have to make any changes. Cluster mode is set to false by default. Can someone with the non cluster mode test this branch?

```python
    if str_to_bool(self.config['cluster']):
            self.collect_cluster(server, device, NaServer)
        else:
            self.collect_vfiler(server, device, NaServer)
```

shortdudey123:
Can you add tests to this for the cluster code you added?

PR #45, opened 2015-01-02
allow to config whether to propagate log of ArchiveHandler to its parent...

hvnsweeting:
... logger

ArchiveHandler is implemented by utilizing logging module, though it is not
considered as a logger. As a handler, it should not log the metrics that
it handles (same as all other handlers do).
This change turns off logging by default and
adds ability to turn that on through its configuration.

### DIFF
--- a/src/diamond/handler/archive.py
+++ b/src/diamond/handler/archive.py
@@ -24,6 +24,7 @@ class ArchiveHandler(Handler):
         # Create Archive Logger
         self.archive = logging.getLogger('archive')
         self.archive.setLevel(logging.DEBUG)
+        self.archive.propagate = self.config['propagate']
         # Create Archive Log Formatter
         formatter = logging.Formatter('%(message)s')
         # Create Archive Log Handler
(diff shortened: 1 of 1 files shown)


hvnsweeting:
this worked for old version, not 4.0 yet, I need to test more

kormoc:
Doesn't the debug handler do this already? Is there a reason to add the functionality to archive handler?

hvnsweeting:
what did you mean the "debug handler"? a handler in ``/handler`` as ``archivehandler.py`` or what ?

This MR is to remove the logging messages which be sent to another log destination (syslog or what configured in ``root`` logger) when using ``archivehandler.py`` - which sends log to a "archive file".

hvnsweeting:
> this worked for old version, not 4.0 yet, I need to test more

I tested with 4.0, the problem is handler config parsing, not this change of code. It worked when I hardcode the path for log_file (as I don't know how to config handlers anymore, see #51

kormoc:
Ohh I see. Thanks!

PR #46, opened 2015-01-03
Update README.md: point to new python-diamond repo

krbaker:
figured this should be fixed, not sure if there is a better way to make this relative


kormoc:
Thanks!

PR #47, opened 2015-01-03
Config file fixes

jaingaurav:
I started testing Diamond 4.0.9 and I ran in a couple of bugs in parsing the existing diamond config file with the new code. The first patch resolves handlers not loading correctly, while the later adds some path existence checks since the config had an empty path specified.

### DIFF
--- a/src/diamond/server.py
+++ b/src/diamond/server.py
@@ -93,8 +93,8 @@ class Server(object):
             sys.exit(1)
 
         handlers = self.config['server'].get('handlers')
-        handlers = handlers.split(',')
-        handlers = map(str.strip, handlers)
+        if isinstance(handlers, basestring):
+            handlers = [handlers]
 
         # Prevent the Queue Handler from being a normal handler
         if 'diamond.handler.queue.QueueHandler' in handlers:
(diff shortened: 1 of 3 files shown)


src/diamond/server.py
```diff
@@ -93,8 +93,8 @@ def run(self):
             sys.exit(1)
 
         handlers = self.config['server'].get('handlers')
-        handlers = handlers.split(',')
-        handlers = map(str.strip, handlers)
+        if isinstance(handlers, basestring):
+            handlers = [handlers]
```
kormoc: We should still split, no?

jaingaurav: There is no need. If there is a comma, indicating multiple elements, a list will be returned with each of the elements. The comma is thus automatically omitted.

kormoc:
Thanks!

chriscowley:
It seems to me that (at least in my case) it is not descending into my collectors directory. More precisely, I have a file cat /etc/diamond/collectors/DiskSpace.conf which contains:

enabled = True
filesystems = ext2,ext3,ext4,xfs,btrfs

And Diamond is still sending info about my NFS mounts to Graphite.

PR #48, opened 2015-01-03
Graphite handler tunning

jaingaurav:
- Set the default pickle port as 2004
- Use 127.0.0.1 instead of 'graphite' in example config


kormoc:
Thanks!

PR #50, opened 2015-01-04
handler config parsing fix

hvnsweeting:


hvnsweeting:
I also tested with just one Handler, without "," at the end, and it still work.

````
# grep 'handlers = ' /etc/diamond/diamond.conf
handlers = diamond.handler.archive.ArchiveHandler
````

hvnsweeting:
NOTICE that I tested with "configobj==5.0.6", older version may handler the right hand side as a String in very old version and as a list only if there is ``,`` in it. By the way, ``configobj==5.0.6`` comes with diamond 4.0 (installed by pip)

jaingaurav:
I'm surprised that it worked with the "," as per the configobj documentation, this would be interpreted as a string. Please see I resolved the issue in pull request #47. This problem was solved in the same manner in the old code.

hvnsweeting:
@jaingaurav  it is because of:

````
In [1]: 'xasfsda'.split(',')
Out[1]: ['xasfsda']

````

so if a string passed, it will return a list after split. Anw, I'm closing this PR as your fix seems more better.

PR #53, opened 2015-01-04
Fix processresource memory collect

hvnsweeting:
I also update the doc and remove some configs that I think are obsoleted.


kormoc:
You can use assertPublishedMetric without an expected value and it'll only assert if it wasn't published. That should work, no?

hvnsweeting:
after rethought, I think this should using mock data same as all other collectors.

jaingaurav:
ping ping could this please be merged?

jaingaurav:
@mattrobenolt: ping

mattrobenolt:
Is there a reason to not also leave the `ext_memory_info` key for older versions? It seems that it'd gracefully work if both keys are present.

jaingaurav:
That would work too as that is what PR #67 tried to do. I guess you could take in part of that PR followed by this. I can make a clean PR for you if you'd like with all the cherry-picks of each of the authors.

shortdudey123:
No longer needed.

PR #60, opened 2015-01-12
Add more verbose output when a collector fails for some reason

shaunduncan:
I noticed that if a collector fails for some reason (i.e. an unchecked exception) no information is written to the logs about the failure. This would make diagnosing problems difficult, so I've added a bare except to log the failure.

### DIFF
--- a/src/diamond/utils/scheduler.py
+++ b/src/diamond/utils/scheduler.py
@@ -83,7 +83,7 @@ def collector_process(collector, metric_queue, log):
             reload_config = True
             pass
 
-        except:
+        except Exception:
             log.exception('Collector failed!')
             break
 
--- a/src/diamond/utils/scheduler.py
+++ b/src/diamond/utils/scheduler.py
@@ -83,6 +83,10 @@ def collector_process(collector, metric_queue, log):
             reload_config = True
             pass
 
+        except:
+            log.exception('Collector failed!')
+            break
+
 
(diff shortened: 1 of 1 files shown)


jamesmulcahy:
+1, looks good to me!

I didn't know about log.exception() before, so I've learnt something new here too :)

nielsole:
It definitely is better than printing nothing, but now it also prints errors, when there are just `sys.exit(0)` calls. It would probably be better to differentiate between `Exception`-subclasses and everything else(eg: `SystemExit`). Not sure how this would look in practice.

jaingaurav:
I think you just need to change `except:` to `except Exception:` that will ensure that you only catch exceptions that inherit from the base class `Exception`

mattrobenolt:
Absolutely yes to what @jaingaurav said.

src/diamond/utils/scheduler.py
```diff
@@ -83,6 +83,10 @@ def collector_process(collector, metric_queue, log):
             reload_config = True
             pass
 
+        except:
```
mattrobenolt:
```python
except Exception:
```

shaunduncan:
Probably a good call. I've made the change.

josegonzalez:
Anyone have any objections to merging this?

mzupan:
I think this is good :+1:

josegonzalez:
Merged, thanks for the PR!

PR #61, opened 2015-01-13
added ES aliases check

prune998:
Added an option that will send metrics of aliased index only... 

One pattern with ES is to alias indices under a static name. re-indexing will be a matter indexing documents in a new index (ex : my_index-2015011302) then change the alias my_index to reference it (and remove the old reference to my_index-2015011301, for example)

This PR will ensure to use the constant alias name as serie name.

### DIFF
--- a/src/collectors/elasticsearch/elasticsearch.py
+++ b/src/collectors/elasticsearch/elasticsearch.py
@@ -70,7 +70,10 @@ class ElasticSearchCollector(diamond.collector.Collector):
             'stats': "Available stats: \n"
             + " - jvm (JVM information) \n"
             + " - thread_pool (Thread pool information) \n"
-            + " - indices (Individual index stats)\n",
+            + " - indices (Individual index stats)\n"
+            + " - aliases (Individual index stats from aliases)\n"
+            + "     will only gather stats from index that are part\n"
+            + "     of an alias",
             'logstash_mode': "If 'indices' stats are gathered, remove "
             + "the YYYY.MM.DD suffix from the index name "
             + "(e.g. logstash-adm-syslog-2014.01.03) and use that "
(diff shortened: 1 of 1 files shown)


bmhatfield:
Hi, This PR appears to fail tests. Would you mind checking in on the Travis build?

src/collectors/elasticsearch/elasticsearch.py
```diff
@@ -209,11 +212,35 @@ def collect_instance_index_stats(self, host, port, metrics):
...
+                self._index_metrics(metrics, 'indices.%s' % name,
+                                    index['primaries'])
+
+
+
+
             self._index_metrics(metrics, 'indices.%s' % name,
                                 index['primaries'])
 
+
```
shortdudey123: remove extra spacing

src/collectors/elasticsearch/elasticsearch.py
```diff
@@ -88,7 +91,7 @@ def get_default_config(self):
             'port':           9200,
             'instances':      [],
             'path':           'elasticsearch',
-            'stats':          ['jvm', 'thread_pool', 'indices'],
+            'stats':          ['jvm', 'thread_pool', 'indices', 'aliases'],
```
shortdudey123: don't enable by default since this will mean a behavior change

src/collectors/elasticsearch/elasticsearch.py
```diff
@@ -209,11 +212,35 @@ def collect_instance_index_stats(self, host, port, metrics):
             indices = result['indices']
         else:
             return
+        
+        if 'aliases' in self.config['stats']:
+          # remove indices not aliased
+          # aggregate values from aliased indexes
+          aliases={}
```
shortdudey123: declare `aliases` before the if block, then you don't need the `else` or the initial declaration in the `if

shortdudey123:
This collector has had a major rewrite since this PR was opened.  Please create a new PR using the current master.  If you revisit this, it would be better to use a Boolean config option instead of this implementation.

PR #62, opened 2015-01-14
nginx: small cleanup - don't define the same thing twice

sanga:


mattrobenolt:
Thanks. :cake: :sparkles:

PR #64, opened 2015-01-15
bugfix: this can be func or data attr depends on psutil ver

hvnsweeting:


mzupan:
:100: thanks for the PR

PR #67, opened 2015-01-16
updated default_info_keys to reflect change in psutil 2.0

prune998:
according to http://grodola.blogspot.com/2014/01/psutil-20-porting.html psutil does not report ext_memory_info starting at version 2.0. Instead, use memory_info_ex.
Adding them both to the default_info_keys ensure memory values will be reported whatever version is used (even if the metrics name changes)


hvnsweeting:
there is also a PR for it here https://github.com/python-diamond/Diamond/pull/53

prune998:
maybe it should be merged then ?

hvnsweeting:
yes, I'm waiting for author

shortdudey123:
No longer needed.

PR #68, opened 2015-01-20
WIP: solr collector modifications

erezzarum:
Allow different webapp context (default is "solr")
Not all Solr installations have all Query Caches enabled by default, we need to validate that those Caches are enabled.
Solr 4.x changes:
- new LotsOfCores feature, make sure that we don't load transient cores that are not loaded
- Deprecated "standard" handler name in favor of "/select"

### DIFF
--- a/src/collectors/solr/solr.py
+++ b/src/collectors/solr/solr.py
@@ -29,6 +29,7 @@ class SolrCollector(diamond.collector.Collector):
         config_help.update({
             'host': "",
             'port': "",
+            'context': "",
             'core': "Which core info should collect (default: all cores)",
             'stats': "Available stats: \n"
             " - core (Core stats)\n"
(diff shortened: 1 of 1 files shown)


shortdudey123:
Can you rebase on master?

src/collectors/solr/solr.py
```diff
@@ -67,8 +69,9 @@ def _try_convert(self, value):
             return value
 
     def _get(self, path):
-        url = 'http://%s:%i/%s' % (
-            self.config['host'], int(self.config['port']), path)
+        url = 'http://%s:%i/%s%s' % (
```
shortdudey123, line 72: Need to account for path.lstrip that is in master.

src/collectors/solr/solr.py
```diff
@@ -138,7 +152,12 @@ def collect(self):
                 ])
 
             if 'query' in self.config['stats']:
-                standard = stats["QUERYHANDLER"]["standard"]["stats"]
```
shortdudey123, line 141: Can you revert this change? #77 handles it better(or bring in those changes).

PR #72, opened 2015-01-20
Make DiskSpaceCollector honor exclude_filters

### LINKED ISSUE #69: DiskSpaceCollector does not honour exclude_filters unless using /proc/mount
This config option is ignored unless the mounts are found in /proc/mounts. If psutil is used to determine the mounts, then no filtering occurs.

I have a fix for this which I'll push shortly.

jamesmulcahy:
Fixes #69


josegonzalez:
This fails tests, can you fix them up? I'll merge once thats done :)

PR #73, opened 2015-01-21
Changes is src/collectors/solr

piontas:
Changed "standard" key in "QUERYHANDLER" to "/select" because in newer version of Solr (4.7.x, etc.), key 'standard' doesn't exist. I've also added "/replication" key which allows to get some replication statistics.


erezzarum:
Hey,
I have already made some modifications to this collector to support 4.x yet to keep the 3.x support.
I have a WIP of making it multi instance aware on the same server, i took the same approach the elasticsearch collector is doing.
https://github.com/python-diamond/Diamond/pull/68

josegonzalez:
Closing in favor of #68

PR #74, opened 2015-01-21
RabbitMQ: collect overall message metrics

russss:
This adds health.[messages|messages_ready|messages_unacknowledged] metrics to the RabbitMQ collector, for monitoring the total number of messages in RabbitMQ.

If you have lots of RabbitMQ queues then pushing per-queue metrics and then summing them isn't ideal.


josegonzalez:
This needs to be rebased in order to be merged, can you do that and add a test if possible?

russss:
Looking at this again, it seems like these metrics were already published by diamond and I just didn't spot them the first time. Sorry for the hassle!

PR #77, opened 2015-01-22
Fix incorrect url build in _get

LannyRipple:
_get incorrectly builds the Solr url including a '/' between
host and path.  All path entries in code begin with '/' so
two show up in the eventual url.  Unit tests missed this since
they do not hit an HTTP server (which admittedly is integration
testing).


shawnbutts:
Since the SolrCollector seems to broken without this, how can we get this merged?

josegonzalez:
Tests need to pass (and this needs a rebase).

shortdudey123:
Can you rebase on master?  The get URL issue was already fixed and so was test.py but I still want to bring in the other changes.

PR #86, opened 2015-01-29
Fixes re-building of SRPM with mock

robinbowes:
The SRPM generated by setuptools does not rebuild under mock (complains about not being able to find python).

This change fixes that.


mzupan:
:+1:

PR #89, opened 2015-01-29
Add HttpJSONPostHandler

shawnbutts:
Post JSON formatted metrics to an HTTP endpoint


josegonzalez:
Fancy. Any reason why this can't be part of the existing HttpHandler?

shawnbutts:
Based on the existing Graphite/GraphitePickle split, I separated it out so there would be less confusion about what to expect on the wire.

PR #90, opened 2015-02-03
Rpm preinstall script

robinbowes:
A couple of rpm-related updates:

1. Add a pre-install script to create a diamond group and user
2. Fix a typo in the rpm pre-uninstall config entry

R.


mzupan:
works good.. thanks for the PR

PR #93, opened 2015-02-04
add trim_backlog support to HttpPostHandler

shawnbutts:
I see little point in letting it build up forever.


josegonzalez:
Tests here as well, though re: #94, we should probably make this a bit generic.

PR #94, opened 2015-02-04
fix trim_backlog for GraphitePickleHandler

shawnbutts:
The GraphiteHandler takes care of keeping self.metrics from growing to large.  But, the this handler also use self.batch that wasn't being shortened

### DIFF
--- a/src/collectors/slony/slony.py
+++ b/src/collectors/slony/slony.py
@@ -1,0 +1,145 @@
+# coding=utf-8
+
+"""
+Collect slony metrics from postgresql
+
+#### Dependencies
+
+ * psycopg2
+
+#### Example Configuration
+
+```
+enabled = True
+
+host = localhost
+port = 5432
+slony_node_string = Node [0-9] - [_a-z0-9]*@(.*).example.com
+
+[instances]
+
+[[database1]]
+slony_db = postgres
+slony_schema = _slony
+
+
+[[database2]]
+user = postgres
+password = postgres
+slony_db = data_db
+slony_node_string = Node [0-9] - [_a-z0-9]*@(.*).i.example.com
+slony_schema = _data_db
+```
+
+"""
+
+import diamond.collector
+
+try:
+    import psycopg2
+    import psycopg2.extensions
+    psycopg2  # workaround for pyflakes issue #13
+except ImportError:
+    psycopg2 = None
+
+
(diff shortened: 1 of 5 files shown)


jaingaurav:
Isn't the backlog already handled in the call to self._send() which calls _send() in the GraphiteHandler base class?

shawnbutts:
For self.metrics list yes.  But, the the picket version also has a self.batch that I could not find being trimmed anywhere.

jaingaurav:
I see, wouldn't it make sense to change the call `self.metrics = [self._pickle_batch()]` to instead append to self.metrics and then flush out `self.batch`? That way you won't have to duplicate the backlog code.

jaingaurav:
also based on you PR for the HTTPPostHandler, it seems like it would be best to make backlog management a more generic functionality for any handler.

shawnbutts:
Good point.  But, I don't think Graphite can accept a list of pickled objects.

shawnbutts:
I agree that it would be best to make this more generic.  These PRs are mostly quick fixes for production issues and the current patterns seems to be to implement specific to the handler.

If I can find the time I'd be willing to look into making backlog management generic. 

Thanks

jaingaurav:
From my rudimentary reading of the code, I believe the problem is that `self.batch` is used in the GraphitePickleHandler instead of plain old `self.metrics`. If you move to using `self.metrics` I think you will just have to create something like a `self._prepare_data` function in the GraphiteHandler which either returns the metrics and a string or pickled data depending on the class.

I think if you take that approach you are likely to remove more code than add much.

shawnbutts:
I'm generally not a fan of moving functionality out of a child class up to the parent if the parent doesn't use it. To me the only time it makes since to do that is when collapsing the child back into the parent.   Here, GraphiteHandler has no use for pickle support otherwise.

However, if that is the recommended way of handling these cases in this project, I'll update the PR when I have time.

josegonzalez:
@shawnbutts have you had a chance to revisit this? Also, will need a test so that we don't break functionality in the future.

shawnbutts:
Hopefully, I'll have time in a week or so to revisit this and #93

rafikk:
Any update on merging in #93 and #94 or other solutions to this problem? This has caused production problems for us.

PR #95, opened 2015-02-05
Added support for cciss device

pb-it:


josegonzalez:
If possible, can you add a test for this?

jaingaurav:
@pb-it Any chance you can add a test case?

PR #97, opened 2015-02-09
Add portstat collector

Nihn:
@mwicat is contributor of this collector.

### DIFF
--- a/src/collectors/portstat/portstat.py
+++ b/src/collectors/portstat/portstat.py
@@ -1,0 +1,69 @@
+
+"""
+The PortStatCollector collects metrics about ports listed in config file.
+
+##### Dependencies
+
+* psutil
+
+"""
+
+from collections import Counter
+
+import psutil
+import diamond.collector
+
+
+def get_port_stats(port):
+    '''
+    Iterate over connections and counts states for specified port
+    :param port: port which stats are collected
+    :return: Counter with port states
+    '''
+    cnts = Counter()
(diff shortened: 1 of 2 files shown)


src/collectors/portstat/portstat.py
```diff
@@ -0,0 +1,69 @@
...
+
+from collections import Counter
+
+import psutil
+import diamond.collector
+
+
+def get_port_stats(port):
+    '''
+    Iterate over connections and counts states for specified port
```
deejay1: language: "*Iterates* over connections" or "Iterate over connections and *count*"

src/collectors/portstat/portstat.py
```diff
@@ -0,0 +1,69 @@
...
+    :param port: port which stats are collected
```
deejay1: "port *for which* stats are collected"

src/collectors/portstat/portstat.py
```diff
@@ -0,0 +1,69 @@
...
+
+"""
+
+from collections import Counter
+
+import psutil
+import diamond.collector
+
+
+def get_port_stats(port):
```
deejay1: Wouldn't it be cleaner to "hide" that inside the collector class? Just asking ;)

Nihn:
ping :).

josegonzalez:
@deejay1 any objections to merging?

deejay1:
@josegonzalez :+1:

PR #102, opened 2015-02-15
Update Makefile so RPM will have python-configobj as a dependancy, this ...

leader716:
The only change in the PR is to include the python-configobj package as a dependency when building the rpm. Currently you can build the RPM, install it and you will only be notified when you attempt to start the service that you need this package installed.


deejay1:
:+1:

mzupan:
:100:

PR #103, opened 2015-02-16
trend_emic: New collector

randomvariable:
This could be too esoteric for inclusion, but the following collector adds support for a type of interface for 3-phase electricity meters seen in some installations.

### DIFF
--- a/src/collectors/trend_emic/trend_emic.py
+++ b/src/collectors/trend_emic/trend_emic.py
@@ -1,0 +1,154 @@
+# coding=utf-8
+
+"""
+The TrendEmicCollector is designed for collecting data from Trend Ethernet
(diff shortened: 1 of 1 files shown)


mattrobenolt:
I'm a -1 since this means someone will need to maintain this collector and pretty sure nobody is even capable of doing that. Also no tests to even begin asserting that it's correct behavior.

randomvariable:
That's fine. Withdrawing.

mzupan:
maybe have a wiki entry for 3rd party collectors like this and have some easy way to install them.

randomvariable:
Might be nice to support multiple entries for "collectors_path", so you could store custom collectors in /etc/diamond/collector_modules, for example.

PR #104, opened 2015-02-16
Implement Basic-Auth for jolokia collector.

mfriedenhagen:
* All our jolokia instances are protected by basic auth normally.
* To be backwards-compatible, username and password are None,
  when both are set, the request is sending credentials pre-
  emptively.

### DIFF
--- a/src/collectors/jolokia/jolokia.py
+++ b/src/collectors/jolokia/jolokia.py
@@ -49,6 +49,7 @@ an mbean.
 """
 
 import diamond.collector
+import base64
 import json
 import re
 import urllib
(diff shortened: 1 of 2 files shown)


josegonzalez:
This seems reasonable, does anyone have any objections to merging?

caskolkm:
Any chance of merging this?

jaingaurav:
@mfriedenhagen I can merge this in. Could you please try rebasing to avoid conflicts. Also, if you could squash at least the first 2 commits so that we don't have that interim pep8 failure

jaingaurav:
Thanks, merged.

PR #105, opened 2015-02-17
add a collector to get ceph usage stats from ceph cluster

psych0d0g:
...l as per pool


josegonzalez:
Can you write a test for this collector?

shortdudey123:
@psych0d0g can you fix the pep8 error and squash your commits?

PR #109, opened 2015-02-20
Setting a precision for publish for sub MS ping times

mzupan:
I needed to test ping times from servers to memcache servers in
SoftLayers network. We’d get > 50ms ping times randomly and bids would
fail taking too long to connect to memcache.

It’s possible to have sub MS response times as seen

```
64 bytes from memcache101.domain.com (10.102.74.173): icmp_seq=1 ttl=64
time=0.157 ms
64 bytes from memcache101.domain.com (10.102.74.173): icmp_seq=2 ttl=64
time=0.148 ms
64 bytes from memcache101.domain.com (10.102.74.173): icmp_seq=3 ttl=64
time=0.155 ms
64 bytes from memcache101.domain.com (10.102.74.173): icmp_seq=4 ttl=64
time=0.124 ms
```

without the precision in there it would send metric to graphite like

0.0000

### DIFF
--- a/src/collectors/ping/ping.py
+++ b/src/collectors/ping/ping.py
@@ -80,4 +80,4 @@ class PingCollector(diamond.collector.ProcessCollector):
                 else:
                     metric_value = 10000
 
-                self.publish(metric_name, metric_value)
+                self.publish(metric_name, metric_value, precision=4)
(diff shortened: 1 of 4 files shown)


josegonzalez:
What was the precision previously? If it wasn't 4, can you make the previous version a default and add a config option for precision to the collector?

mzupan:
precision=0 by default. The issue is any ping under 1ms becomes 0.0 due to it. All the tests still work so I don't see a need to set that as a feature. It'll just lead to issues where people over look that setting

jaingaurav:
@mzupan: I think we can merge this PR in. Could you please squash the commits before we do so?

shortdudey123:
I would prefer to use #124 instead.

PR #110, opened 2015-02-20
Fix  #108

### LINKED ISSUE #108: memory collectors uses wrong property to get actual memory available
According to documentation (http://pythonhosted.org/psutil/#psutil.virtual_memory) *free* isn't the right property to read amount of available memory - https://github.com/python-diamond/Diamond/blob/master/src/collectors/memory/memory.py#L115

```
free: memory not being used at all (zeroed) that is readily available; note that this doesn’t reflect the actual memory available (use ‘available’ instead).
```

Nihn:
Fix #108


deejay1:
:+1:

mzupan:
looks good.. thanks for the PR

chriscannon:
How do you enable MemAvailable? Seems like I need to turn on detailed somewhere?

PR #112, opened 2015-02-26
[Postgres] Added more debug log

lyrixx:
because, if the database connection can not be established,
there is no error, and so it's really not easy to detect.


mattrobenolt:
Thanks! :cake:

PR #117, opened 2015-03-02
[Rabbitmq] Added a way to replace . character in queue and vhosts name

lyrixx:
Because in our infrastructure, queues and vhosts have "." in their name.
And with graphite, a "." is a directory separator.
So it becomes very hard to do globbing and configurable dashboards.

### DIFF
--- a/src/collectors/rabbitmq/rabbitmq.py
+++ b/src/collectors/rabbitmq/rabbitmq.py
@@ -82,6 +82,8 @@ class RabbitMQCollector(diamond.collector.Collector):
             'host': 'Hostname and port to collect from',
             'user': 'Username',
             'password': 'Password',
+            'replace_dot':
+            'A value to replace dot in queue names and vhosts names by',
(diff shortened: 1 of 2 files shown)


beeftornado:
+1 for this feature

josegonzalez:
Can you add a test for this feature?

lyrixx:
@josegonzalez Sorry, I'm a bit new with python; and I think I will not be able to add test... Could you do it for me? or could you explain how to do it?

beeftornado:
I can volunteer to do the test tonight.

beeftornado:
I wasn't sure what the best workflow would be for this situation, but I did what I thought was best.

@lyrixx I forked your version and added the test and submitted a pull request to you. I *think* if you accept it in the same branch this PR is for then it should show up here.

lyrixx:
@beeftornado What you did is perfect. Thanks a lot ;)

lyrixx:
ping ?

lyrixx:
pong ;)

PR #118, opened 2015-03-04
KafkaCollector > Fix the key_prefix issue

kngjle:
It do not parse rightly the objectname that have more than 2 attributes.
For example,
kafka.network:type=RequestMetrics,name=RequestsPerSec,request=Produce
=> RequestMetrics.RequestsPerSec,request.RequestsPerSec,request

It seems better to be parsed like this.
=> RequestMetrics.RequestsPerSec.Produce

I have defined the function that build the key_prefix as above.


shortdudey123:

PR #120, opened 2015-03-05
Add collector for twemproxy aka nutcracker.

gbock:
Collector for twemproxy: https://github.com/twitter/twemproxy


gbock:
Closing, I'll open a new one as my git foo is bad.

PR #124, opened 2015-03-06
Updated PingCollector to work with v4 also v3

mzupan:
This is an updated ping collector to work with v4 forking but still work with v3 way of doing things. 

It adds two new config options

```
extended: (true|false)
count: number
```

It also changes the old format of `target_key = host` to just `target = host` 

If extended is enabled it will log the following

```
hostname.min: val
hostname.max: val
hostname.avg: val
hostname.mdev: val
```

A lot of times a single ping tells you nothing but the min/max/avg of 20-30 pings might be more telling. 

It's made to support the new forking by creating different configs with a space in the config to fork off a new process. It will also support the old way of doing it all in a single config.

### DIFF
--- a/src/collectors/ping/ping.py
+++ b/src/collectors/ping/ping.py
@@ -16,9 +16,13 @@ Create a file named: PingCollector.conf in the collectors_config_path
 
  * enabled = true
  * interval = 60
- * target_1 = example.org
- * target_fw = 192.168.0.1
- * target_localhost = localhost
+ * target = example.org
+ * extended = true
+ * count = 10
+
+The extended and count are optional. If enabled extended will show
+min/max/avg/mdev for the results. Count will do more then a single
+ping to get those results.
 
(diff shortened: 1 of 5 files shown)


mzupan:
not sure why the tests are failing.. it was working on my laptop.. I'll look into it

mzupan:
ok tests are working now

mattrobenolt:
Isn't this something you'd use statsd for to get the aggregates? Personally, I don't think this is a thing that Diamond should do. There are other tools that ingest all the raw stats and aggregate them over periods of time.

jaingaurav:
I think that collectors should really only be reporting raw values, unless the source that it is reporting from already provides computed values. For example the disk S.M.A.R.T. collector may report max values since those are reported by the disk firmware itself.

The disk stats collector is an example of a collector that computes a lot of aggregates and as a result reports a lot of duplicate information and is very confusing to understand.

PR #126, opened 2015-03-11
Add DiskTemperatureCollector which only extracts the disk temp from S.M.A.R.T.

jaingaurav:
This collector uses hddtemp to collect only the disk temperature from the disk S.M.A.R.T information. This can be faster than smartctl since it only extracts a single value.


jaingaurav:
Done

deejay1:
:+1:

jaingaurav:
@mattrobenolt @josegonzalez could this please be merged?

PR #128, opened 2015-03-11
ipmisensor.py was parsing sensor names with space wrong

sirtux:
ipmitool is having sensor names with space, eg "01-Inlet Ambient" (at least on HP ilo ipmi implementation), which caused with the current source code splits in the path, this has been fixed now.

### DIFF
--- a/src/collectors/ipmisensor/ipmisensor.py
+++ b/src/collectors/ipmisensor/ipmisensor.py
@@ -27,7 +27,8 @@ class IPMISensorCollector(diamond.collector.Collector):
             'bin': 'Path to the ipmitool binary',
             'use_sudo': 'Use sudo?',
             'sudo_cmd': 'Path to sudo',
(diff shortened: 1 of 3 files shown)


jaingaurav:
This behavior is intentional. By splitting the source code paths it allows grouping of metrics for the same component.

sirtux:
Hi,

this works well for some implementations of ipmi, but for the HP variant this is more of a hazzle.
I was just thinking about making this behaviour configurable, and retain the current state as default,
what do you think?

thanks
Tom

jaingaurav:
That seems like a reasonable suggestion. You could add a config value such as 'delimiter' which could be set to values such as '.' or '_'.

Could you also include the output of your ipmitool as one of the test fixtures? It would be easier to understand then, and ensure that future changes respect the variations in ipmi output.

sirtux:
I will clear that on a new pull request in some days after cleaning up the mess I created and close this request for now.

PR #130, opened 2015-03-12
Added delimiter option to ipmisensor.py

sirtux:
Added an option to impisensor.py for having a flexible delimiter configuration, defaulting to '.', which is the current default.
Change runs through the current tests, but a test case for a dynamic delimiter has to be written, but I have no idea howto implement that.


jaingaurav:
This looks great. However could you add a test case which includes sample output from the HP IPMI?

sirtux:
How do I add a second test case? All tests I have seen so far only have one fixture - can I simply create a second test file?

jaingaurav:
Just duplicate the test_should_work_with_real_data function in testipmisensor.py within the same file and change the getFixture('ipmitool.out') call to return your file instead. For further reference, you can look at some other test case files such as testsmart.py.

sirtux:
Here we go with the added tests

jaingaurav:
:+1: :+1: :+1: :+1:

Thanks a lot! This looks really good. The test cases really help provide a basis for this change.

sirtux:
Hi all, do you have an idea when this request will be merged? I would like to go back to create packages for us from the master branch, and no longer our own...

Or is anything missing?

Thanks,
Tom

jaingaurav:
Code has been merged. Thanks for the work and I apologize regarding the late merge.

PR #131, opened 2015-03-13
[portstat] fix missing 'PortStatCollector' in get_port_stats method call

Nihn:


mzupan:
@Nihn looks like we might also need to add self into the function above too

Nihn:
What do you mean? I don't see where :).

jaingaurav:
Did you mean the get_port_stats function signature? If so, I don't think it needs it because its a static method.

mattrobenolt:
`self` isn't right. It's a static method.

If anything, switch it to `PortStatCollector.get_port_stats(port)`.

mattrobenolt:
OR, just move `get_port_stats` off of the class entirely and just use it like a normal function instead of a method.

Nihn:
self is 100% valid but ill do as you wish :).

mattrobenolt:
> self is 100% valid but ill do as you wish :).

I didn't say it wasn't valid, but it's not right. :)

mattrobenolt:
thx :cake: :sparkles:

PR #142, opened 2015-03-27
Initial version of SignalFx plugin

cep21:
SignalFx is a metrics and monitoring service.  This plugin allows Diamond users to send metrics to us from Diamond.  Learn more about SignalFx at https://signalfx.com/


baris:
It would be really nice to have this handler included with the Diamond. What's the status on this PR? I'd be happy to volunteer time on testing. I've already tested the handler in a development environment and verified that it's working as expected.

mzupan:
setup a test account and does seem to be working

:+1:

PR #147, opened 2015-04-01
Add IPv6 support to IPCollector

russss:
The /proc/net/snmp file only contains IPv4 metrics. This change additionally reads IPv6 metrics from /proc/net/snmp6.

These metrics are all named with a "6" in them so I'm just putting them in the same namespace as the v4 metrics.


src/collectors/ip/test/testip.py
```diff
@@ -38,49 +38,70 @@ def test_import(self):
     def test_should_open_proc_net_snmp(self, publish_mock, open_mock):
         IPCollector.PROC = ['/proc/net/snmp']
         open_mock.return_value = StringIO('')
-        self.collector.collect()
+        self.collector.collect_ipv4()
```
jaingaurav: I think it makes more send to call `self.collector.collect()` instead of the ipv4/ipv6 variants. I'm sure you can configure different mock return values based on the filename passed in.

src/collectors/ip/ip.py
```diff
@@ -56,30 +69,38 @@ def get_default_config(self):
...
         config.update({
             'path': 'ip',
-            'allowed_names': 'InAddrErrors, InDelivers, InDiscards, ' +
-            'InHdrErrors, InReceives, InUnknownProtos, OutDiscards, ' +
-            'OutNoRoutes, OutRequests'
+            'allowed_names': ','.join(self.DEFAULT_METRICS)
         })
         return config
 
+    def open_file(self, filepath):
```
jaingaurav: does this really need to be abstracted out into a separate function? Can'e you directly issue an open and catch the permission exception?

jaingaurav:
The code looks great. Made a few minor comments. After the changes, we should be able to merge the code.

jaingaurav:
@russss Any update on this PR?

PR #148, opened 2015-04-01
influxdb version 0.9 support

imreACTmd:
Signed-off-by: imre Fitos imre@act.md

### DIFF
--- a/src/diamond/handler/influxdbHandler.py
+++ b/src/diamond/handler/influxdbHandler.py
@@ -29,6 +29,7 @@ username = root
 password = root
 database = graphite
 time_precision = s
+influxdb_version = 0.8
 """
 
 import time
@@ -71,6 +72,7 @@ class InfluxdbHandler(Handler):
         self.metric_max_cache = int(self.config['cache_size'])
         self.batch_count = 0
         self.time_precision = self.config['time_precision']
+        self.influxdb_version = self.config['influxdb_version']
 
         # Initialize Data
         self.batch = {}
@@ -100,6 +102,7 @@ class InfluxdbHandler(Handler):
             'database': 'Database name',
             'time_precision': 'time precision in second(s), milisecond(ms) or '
             'microsecond (u)',
+            'influxdb_version': 'InfluxDB API version, default 0.8',
         })
 
         return config
@@ -120,6 +123,7 @@ class InfluxdbHandler(Handler):
             'batch_size': 1,
             'cache_size': 20000,
             'time_precision': 's',
+            'influxdb_version': 0.8,
         })
 
         return config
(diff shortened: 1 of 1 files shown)


imreACTmd:
Sure, if that's what holding it back from accepting the pull request I will replace the elif with an else, just let me know.

src/diamond/handler/influxdbHandler.py
```diff
@@ -120,6 +123,7 @@ def get_default_config(self):
             'batch_size': 1,
             'cache_size': 20000,
             'time_precision': 's',
+            'influxdb_version': 0.8,
```
mattrobenolt: Shouldn't this be a string? Version numbers aren't floats.

mattrobenolt:
I agree with @ctrlrsf, I don't want to be revisiting this when 0.10 or something is released. So instead of passing along version, what about just a boolean for old/new behavior?

ctrlrsf:
@mattrobenolt I checked with folks on Influx IRC channel and turns out API could still continue to evolve in future versions (duh). Boolean I think makes sense now, but will we then end up with `influx_compat_v08` and `influx_compat_v09` variables if there's another breaking change?

Another alternative was maybe calling it `influx_compat_mode` and it could then be set to "v08" now to support 0.8.x server, and left unset to support latest (0.9.x). Same variable could later be set to "v09" to support 0.9.x server if 0.10.0 changes something else.

Investocat:
There isn't a way to query the api version from the server itself?

imreACTmd:
Looks like there is an X-Influxdb-Version header in 0.8 and 0.9, I'll see what I can do.

airtoxin:
It needs to change import statement because currently influxdb-python modules uses 0.9.x api based client by default.

```python
 try:
    if self.influxdb_version == '0.8':
         from influxdb.influxdb08 import InfluxDBClient
    else:
         from influxdb import InfluxDBClient
 except ImportError:
     InfluxDBClient = None
```

jwestboston:
+1

justin8:
This doesn't seem to work for me? The latest influxdb lib apparently needs to import influxdb.influxdb08 for compatibility since the api changed. but I can't see how this code would handle that at all. Perhaps after establishing the desired version, import the correct version of InfluxDBClient?

imreACTmd:
I will have an updated version soon! Waiting for the 0.9 API to stabilize, they just introduced a new plaintext API as well.

justin8:
Good plan. I'll stay subscribed here then. Thanks

jwestboston:
Looking forward to this now that InfluxDB 0.9.0 is released as stable.

slnowak:
What's the status of this PR? Could you merge it?

imreACTmd:
I'm back from vacation and will be updating it soon (unless someone else wants to do it)

jwestboston:
Merge merge merge :-)

Ormod:
Just a FYI In order to get legible results from Diamond through to Influxdb 0.9.3 (influxdb-python 2.8) and from there to view them reasonably in Grafana 2.1 I had to also add this commit: https://github.com/Ormod/Diamond/commit/dbe971aaf3df8809def408468928bae0893eaa65

rhunwicks:
We're waiting for this PR to merge before deploying Diamond.

src/diamond/handler/influxdbHandler.py
```diff
@@ -170,11 +174,19 @@ def _send(self):
...
+                    if self.influxdb_version == "0.8":
+                        for path in self.batch:
+                            metrics.append({
+                                "points": self.batch[path],
+                                "name": path,
+                                "columns": ["time", "value"]})
+                    elif self.influxdb_version == "0.9":
+                        for path in self.batch:
+                            metrics.append({
+                                "name": path,
```
peggyl: This key name should be "measurement", not "name".

This is required to be consistent with influxdb-client: https://github.com/influxdb/influxdb-python/blob/efdac997faeb36984fa4c949e1c57ca5a6e1303d/influxdb/line_protocol.py#L94

jaingaurav:
@imreACTmd Any chance you can resist this PR and address the comments?

imreACTmd:
Hi Gaurav, I will definitely try!

dibits:
Tested it with the recommended changes and it works like a charm. If @imreACTmd can't do it soon I'll send another pull request with the changes.

shortdudey123:

PR #149, opened 2015-04-06
jolokia: regex bean matching/renaming;  files: allow upper case metrics

sbrynen:
Jolokia Collector:
  allows bean matching on regex
  allows renaming of returned metrics
  removes potentially bad naming in metrics (like double dots)

Files Collector:
  allows upper case metric names, and negative values

### DIFF
--- a/src/collectors/files/files.py
+++ b/src/collectors/files/files.py
@@ -11,7 +11,7 @@ import diamond.collector
 import os
 import re
 
-_RE = re.compile(r'([a-z0-9._-]+)[\s=:]([0-9-]+)(\.?\d*)')
+_RE = re.compile(r'([A-Za-z0-9._-]+)[\s=:]+(-?[0-9]+)(\.?\d*)')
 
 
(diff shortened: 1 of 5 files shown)


src/collectors/files/files.py
```diff
@@ -11,7 +11,7 @@
 import os
 import re
 
-_RE = re.compile(r'([a-z0-9._-]+)[\s=:]([0-9-]+)(\.?\d*)')
+_RE = re.compile(r'([A-Za-z0-9._-]+)[\s=:]+(-?[0-9]+)(\.?\d*)')
```
mattrobenolt: Why is this being changed?

sbrynen: As noted, the files collector (which I also wrote) wouldn't accept an upper case metric name and/or a negative value.  This is fixes as well (see note above in initial commit):
Files Collector:
allows upper case metric names, and negative values

mattrobenolt:
noidea

PR #151, opened 2015-04-16
Add nvidia-smi collector

janisz:
Simple collector using `nvidia-smi` to grab data about GPU state


shortdudey123:
https://github.com/python-diamond/Diamond/pull/370 will supercede this

PR #152, opened 2015-04-16
Fix portstat tests

patricklucas:
These were broken by the refactor in 4234986c.


patricklucas:
Looks like you fixed this a different way four days after I posted this.

PR #160, opened 2015-04-20
multiple instance support for postgres collector

himanchali:
If more than one server is running on same host on different port , current collector doesn't support it. This pull request is to handle the condition where multiple db servers running on single host.


mzupan:
for v4 of diamond it will support it.. If you config is like

PostgreSQL.conf then your 2nd instance would be "PostgreSQL other.conf"

the space means load another instance of the collector

PR #163, opened 2015-04-22
Update setup.py

jumping:
According to the document(https://docs.python.org/2/library/platform.html), the dist function deprecated since version 2.6.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,8 +42,8 @@ else:
         ('share/diamond/user_scripts', []),
     ]
 
-    distro = platform.dist()[0]
-    distro_major_version = platform.dist()[1].split('.')[0]
+    distro = platform.linux_distribution()[0]
+    distro_major_version = platform.linux_distribution()[1].split('.')[0]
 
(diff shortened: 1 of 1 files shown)


mattrobenolt:
But `linux_distribution` was just added in 2.6. I believe we still attempt to maintain compatibility with 2.5.

I understand it's deprecated, but has it been removed and is it causing issues with more modern setups?

If anything, I'd be open to a fallback strategy here, where it tries to use `linux_distribution()`, but fails back to `dist()` on `AttributeError` or whatever.

jumping:
Yes, may could not change that now.

mattrobenolt:
fwiw, `dist()` still works just fine in python 3.4, so I see no reason to change this.

```
$ python3
Python 3.4.0 (default, Apr 11 2014, 13:05:11)
[GCC 4.8.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import platform
>>> platform.dist()
('Ubuntu', '14.04', 'trusty')
```

PR #164, opened 2015-04-22
added logrotate for these logs under the /var/log/diamond

jumping:


mattrobenolt:
@jumping why did you close this?

mattrobenolt:
Also, I'm not entire sure this will work. logrotate would also need to send a signal to diamond to drop the old file descriptor and pick up a new one. And Diamond would need to support this behavior. Can be added though because logrotate should be a thing we do.

mattrobenolt:
fwiw, Diamond supports it's own log rotation with `handlers.TimedRotatingFileHandler`. Though not as versatile as actual logrotate, it does the job and I don't see any immediate reasons to switching unless a pull request covered the scenario I mentioned above.

jumping:
I added this function because the disk volume is very precious for us running instances on AWS. Default the root disk is only 8GB.

mattrobenolt:
Is the built in `handlers.TimedRotatingFileHandler` not sufficient here? I assume you're  not generating multiple GB of logs per day.

jumping:
I just run Diamond on some production instances, so the volume of logs are not sure.

mattrobenolt:
I understand, but you're not answering my question. Are you using the build in `TimedRotatingFileHandler`? That should be rotating logs for you daily.

Even in production (we use it very extensively), it'd be pretty extreme to generate multiple gigs of logs per day. So let me know if the `TimedRotatingFileHandler` is not working or is not enough for you. It's what we use in production just fine and have never had an issue.

jumping:
Another question, does the Diamond support custom ? Now the node on Graphite like as server.ip.cpu.cpu0.iowat, can I change like as  server.tagname.ip.cpu.cpu0.iowait ? In which the tagname  is a value from every instance .

jumping:
Still not use the TimeRotatingFileHandler.

PR #165, opened 2015-04-23
Fix elb_names

bclermont:
turn string into ``,`` separated list

If elb_names is specified it just ``process_elb`` trough every chars that made it's names


bclermont:
https://github.com/python-diamond/Diamond/blob/master/src/collectors/elb/test/testelb.py#L137

the test assume it's already a list

bclermont:
ok I found why

PR #167, opened 2015-04-27
Log exceptions from collectors

joel-airspring:
Here's some code to log exceptions from collector runs.

I'm using diamond from latest pip (4.0.41), and it contains a bug in processresources.py.  However, it was silently failing because collector.py swallowed the exception without logging it.

It's not relevant to this PR, but for the record, the bug I hit in processresources.py was fixed by this commit:  023895b5bea4295e738056e398515e8e273a97b1


mattrobenolt:
Seems reasonable!

joel-airspring:
Awesome.  Thanks!

PR #169, opened 2015-04-27
Add config filename to error messages from parsing collector config file...

joel-airspring:
Supposed you had a syntax error in a collector config file, like:

    enabled = True
    enabled = True

In this case, you get an error message that says
> Duplicate keyword name at line 3.

Good explanation, but it can require looking through dozens of config files to figure out which one is causing the error.

A better error message is produced by this PR:
> Failed to load config file /etc/diamond/collectors/VMStatCollector.conf due to Duplicate keyword name at line 3.

### DIFF
--- a/src/diamond/utils/config.py
+++ b/src/diamond/utils/config.py
@@ -99,7 +99,12 @@ def load_config(configfile):
                 if collector not in config['collectors']:
                     config['collectors'][collector] = configobj.ConfigObj()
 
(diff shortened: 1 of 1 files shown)


jaingaurav:
@joel-airspring lgtm. Could you please squash your commits so that I can merge it all in?

joel-airspring:
Done.

jaingaurav:
Thanks, merged

PR #171, opened 2015-04-28
Escape reserved characters in Jolokia domains

cbeer:
It appears the Jolokia domains need to be escaped with the Jolokia-specific escape sequence, even when used as URL parameters (despite my reading of the documentation..):

https://jolokia.org/reference/html/protocol.html

### DIFF
--- a/src/collectors/jolokia/jolokia.py
+++ b/src/collectors/jolokia/jolokia.py
@@ -153,7 +153,7 @@ class JolokiaCollector(diamond.collector.Collector):
 
     def read_request(self, domain):
         try:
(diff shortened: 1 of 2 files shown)


mattrobenolt:
Can you add a test or multiple tests for this?

cbeer:
I've added some unit tests for the new method. I'm not sure if/how to hook into the request mocking though.

mattrobenolt:
That's good enough for me! :)

Thanks. :cake: :sparkles:

cbeer:
Thanks. I'm also going to start looking into the Jolokia POST request format, which may allow us to avoid this nasty escaping business in the first place (plus allow batch requests..)

PR #174, opened 2015-05-04
add a http flatfile collector to gather key=value metrics from http sources

psych0d0g:
Please have a look and merge if everything is okay


mattrobenolt:
What's the use case for this?

Either way, the tests you added are failing, so check into that.

Lastly, please revert the change to the `debian/changelog` file.

src/collectors/httpflat/httpflat.py
```diff
@@ -0,0 +1,59 @@
...
+    def _flat_to_flat_metrics(self, prefix, data):
+        data = dict(line.split("=") for line in data.split("\n"))
+        for key in data:
+            yield key, data[key]
+
+    def collect(self):
+        url = self.config['url']
+
+        req = urllib2.Request(url)
+        req.add_header('Content-type', 'application/text')
```
mattrobenolt: Is this required for some reason? If not, this is unnecessary.

src/collectors/httpflat/httpflat.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def get_default_config(self):
+        default_config = super(HTTPFLATCollector, self).get_default_config()
+        default_config.update({
+            'path': 'httpflat',
+            'url': 'http://localhost/stat'
+        })
+        return default_config
+
+    def _flat_to_flat_metrics(self, prefix, data):
```
mattrobenolt: This whole thing should just be done as a generator. yielding for each line. Right now, you're splitting the lines upfront, then yielding all the parsed values. Would be better to iterate each line and parse separately.

src/collectors/httpflat/httpflat.py
```diff
@@ -0,0 +1,59 @@
...
+            content = resp.read()
+
+            try:
+                data = content[1:]
+            except ValueError as e:
+                self.log.error("Can't parse JSON object from %s. %s", url, e)
+            else:
+                for metric_name, metric_value in self._flat_to_flat_metrics(
+                        "", data):
+                    self.publish('voip.rtstatng.' + metric_name, metric_value)
```
mattrobenolt: Seems like this prefix is hardcoded? Is this all for a voip service?

psych0d0g:
Closed for now, will redo pull request more cleanly soon

mattrobenolt:
You didn't need to close the pull request to clean up and address feedback. :)

psych0d0g:
i have some more stuff ill rewrite and add thats why i closed it and im going to open a new "clean" one soon

src/collectors/httpflat/httpflat.py
mattrobenolt: Also worth noting, there is no such thing as an `application/text` type. If anything, it'd be `text/plain`, but on a request, this describes the content-type of the body. In this case, you're making a GET, so there is no body, and this isn't needed.

PR #177, opened 2015-05-06
Update jolokia collector to support jolokia behind basic auth

cbeer:
Some Jolokia instances may be protected by basic auth.

Also, add the ability to override Python's default user-agent, as some versions of jolokia key off the user agent to decide whether to use basic auth (e.g. only support Basic Auth if `curl` is asking..).  See https://issues.jboss.org/browse/ENTESB-2418

### DIFF
--- a/src/collectors/jolokia/jolokia.py
+++ b/src/collectors/jolokia/jolokia.py
@@ -53,6 +53,7 @@ import json
 import re
 import urllib
 import urllib2
+import base64
 
 
 class JolokiaCollector(diamond.collector.Collector):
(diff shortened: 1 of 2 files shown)


cbeer:
Do you have any suggestions for testing this? I took the basic auth code from the monit collector, but that doesn't seem to be tested there either..

cbeer:
Please do let me know if there's a better way to test the authentication configuration.

caskolkm:
Any chance of merging this?

jaingaurav:
PR #104 has been merged. Is this diff still needed?

josegonzalez:
Closing as #104 has been merged and no one responded to @jaingaurav.

PR #180, opened 2015-05-07
Add JolokiaBulkCollector to gather metrics using the bulk request syntax

cbeer:
The `JolokiaBulkCollector` uses Jolokia's bulk request syntax [1] to optimize retrieving values from many mbeans. As of Jolokia 1.2.2 (and possibly earlier..), mbeans can be listed using wildcards and, with the bulk request api, can be retrieved directly. 

With the `JolokiaCollector`, when collecting metrics from JVMs with many mbeans, often the collector will hit the timeout while trying to process each mbean (even for generous intervals in excess of 300s). With the bulk collector, the collector is able to retrieve multiple mbean metrics without incurring the same overhead as many individual requests, and is able to finish running in <10s.

This initial work does not implement the mbean regex feature, and requires that all mbeans be explicitly listed (possibly using Jolokia's wildcard syntax). I suspect they could be readded in the future, if desirable. 

This also includes #177, but could be trivially refactored to be stand-alone.

[1] https://jolokia.org/reference/html/protocol.html#post-request

### DIFF
--- a/src/collectors/jolokia/bulk_jolokia.py
+++ b/src/collectors/jolokia/bulk_jolokia.py
@@ -1,0 +1,117 @@
+# coding=utf-8
+
+"""
+ Collects JMX metrics from the Jolokia Agent. Jolokia is an HTTP bridge that
+provides access to JMX MBeans without the need to write Java code. See the
+[Reference Guide](http://www.jolokia.org/reference/html/index.html) for more
+information.
+
+The liste MBean patterns will be queried for metrics. All numerical values will
+be published to Graphite; anything else will be ignored.
+The JolokiaBulkCollector will create a reasonable namespace for each metric
+based on each MBeans domain and name.
+e.g) ```java.lang:name=ParNew,type=GarbageCollector``` would become
+```java.lang.name_ParNew.type_GarbageCollector```.
+
+#### Dependencies
+
+ * Jolokia
+ * A running JVM with Jolokia installed/configured
+
+#### Example Configuration
+
+JolokiaBulkCollector is configured to query specific MBeans by
+providing a list of ```mbeans```. The list of mbeans may include the
+wildcard '*' character, or any other valid Jolokia pattern
+(see https://jolokia.org/reference/html/protocol.html#read).
+
+The ```rewrite``` section provides a way of renaming the data keys before
+it sent out to the handler.  The section consists of pairs of from-to
+regular expressions.  If the resultant name is completely blank, the
+metric is not published, providing a way to exclude specific metrics within
+an mbean.
+
+```
+    host = localhost
+    port = 8778
+    mbeans = "java.lang:*,type=GarbageCollector",
+    [rewrite]
+    java = coffee
+    "-v\d+\.\d+\.\d+" = "-AllVersions"
+    ".*GetS2Activities.*" = ""
+```
+"""
+
+import base64
+import json
+import urllib
+import urllib2
+
+from jolokia import JolokiaCollector
+
+
+class JolokiaBulkCollector(JolokiaCollector):
+    # override to allow setting which percentiles will be collected
+    def get_default_config_help(self):
+        config_help = super(JolokiaBulkCollector,
+                            self).get_default_config_help()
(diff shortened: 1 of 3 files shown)


cbeer:
@mattrobenolt ? Supporting the bulk request syntax is an important feature for my use of Diamond/Jolokia because the single metric version is unacceptable slow (and, seemingly, error prone with ever-changing mbeans like Apache Solr uses).

If you're not interested in merging this, I'll look at alternative means of distributing this collector.

josegonzalez:
@cbeer sorry for leaving this in the queue.

Is there a reason we can't make this the default collector?

cbeer:
@josegonzalez I'm all for it being the default collector, but does that suggest it needs to support the regular-expression style selection and empty-mbeans-means-all-mbeans that the current `JolokiaCollector` does?

cbeer:
bump?

MichaelDoyle:
Hey @cbeer, I think this is awesome. Ideally, in my view, we should merge these two features (basic auth, and bulk query) into the original collector.

I'm interested to know how many beans is "many", in your case. And, is the primary cost savings because of less round trips, or because you are querying far fewer beans now (via whitelist). What would happen if this implementation asked for "all" beans in the bulk query by default? 

The original jolokia collector does a listing of available beans, then it goes domain by domain to fetch the full set of data for each. It doesn't actually do a query per bean, although I can see how this could still be a lot of requests. I don't know how large the result set would be, but we could easily update the original jolokia collector to fetch all beans in 2 requests. One for the listing (to get the list of domains), and a bulk request to fetch all beans for all domains.

The original jolokia collector does all filtering on the collector side. I like how your implementation let's jolokia do the work. I think this makes a lot more sense, especially for a specific whitelist of mbeans. I would also prefer "querying" jolokia via wildcard syntax versus doing regex matches after the fact. I'm not a huge fan of the regex matching as implemented today, although I could see the argument for it if jolokia's wildcards aren't expressive enough. 

Thoughts?

shawnbutts:
This would help with collecting metrics from Cassandra.  In my case "too many beans" is 5289.

brucelee123:
bump?

kullervo16:
Although this request seems to be dangling some time, for those who come here later on... I tested with the bulk collector attached in this pull request.... it is working and it's much more efficient for large collections than the default one. It also handles the discovery of "dynamic" MBeans (which were not found by the default one if not present at first scan).

However, I did encounter some issues :
* when you specify a wrong pattern, no indication -> added errorhandling
* when you specify a "complete" expression that results in direct values, the parsing that expects a key for the domain generates an exception -> always include a wildcard in  your pattern, even if you specify it completely (that way, the path is repeated and the parsing works fine)
* when some attributes are not present in some of the instances, the complete request fails -> added a flag (cfr http://jolokia.963608.n3.nabble.com/MemoryPool-java-lang-UnsupportedOperationException-Usage-threshold-is-not-supported-td3101873.html)
* when the path does not contain a "/" at the end, he won't connect. When it does, the counter name contains the / -> added the "/" only in the URL.

Here are my modifications for the path/error issue :
```
def fetch_response(self):
        path = self.config['path']
        if not(path.endswith('/')):
            path = path + '/'
        url = "http://%s:%s/%s?ignoreErrors=true" % (self.config['host'],
                                   self.config['port'],
                                   path)
      .....
```

This is how to add clear errors when things go wrong (logging imported of course) :
```
def collect(self):
        requests = self.fetch_response()

        for listing in requests:
            if listing['status'] == 200:
                domains = listing['value']
                for domain in domains.keys():
                    if domain not in self.IGNORE_DOMAINS:
                        self.collect_bean(domain, domains[domain])
            else:
                logging.critical("%s(%s) : %s",listing['request'],listing['error_type'],listing['error'])
```

shortdudey123:
@cbeer can you rebase on master, and look at incorporating @MichaelDoyle's and @kullervo16's feedback? thanks!

shortdudey123:
@cbeer can you fix the failing tests? or correct the behavior of the collector?

PR #182, opened 2015-05-09
Adjust staggering of collectors if they are unable to process within their deadline

jaingaurav:
Currently the multi-threaded collector framework staggers collectors to allow for collectors to run at different times. This reduces the load on the system as multiple collectors do not bombard the system at the same time. Unfortunately if a collector is firing close to the end of its interval it is possible that it may miss its deadline.

To fix this, we can reduce the stagger delay anytime the collector misses its window. Eventually, the collector will settle on a value that is safe for its interval and processing time.

Note that this reverts PR #167 it infact broke the handling of signals in the while loop. From my testing I didn't see any problem with collectors reporting exceptions in the log.

### DIFF
--- a/src/diamond/collector.py
+++ b/src/diamond/collector.py
@@ -481,8 +481,6 @@ class Collector(object):
                     metric_name = 'collector_time_ms'
                     metric_value = collector_time
                     self.publish(metric_name, metric_value)
-        except Exception, e:
-            self.log.exception(e)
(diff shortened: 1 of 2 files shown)


jaingaurav:
ping @mattrobenolt @josegonzalez

jaingaurav:
Thanks a lot!

PR #184, opened 2015-05-11
adding sysconfig for init script for RPM 

himanchali:
added sysconfig file for diamond for rpm to control init script to be enabled or disabled. Currently it's always enabled by default and there is no other way to control it. it's needed if we want to use daemon instead of init script.


shortdudey123:
this would break the pid file option in the config file

PR #185, opened 2015-05-11
Fix import statement

airtoxin:
It needs to change import statement because currently influxdb-python modules uses 0.9.x api based client by default.
https://github.com/influxdb/influxdb-python#influxdb-v08x-users

### DIFF
--- a/src/diamond/handler/influxdbHandler.py
+++ b/src/diamond/handler/influxdbHandler.py
@@ -35,7 +35,7 @@ import time
 from Handler import Handler
 
 try:
-    from influxdb.client import InfluxDBClient
+    from influxdb.influxdb08 import InfluxDBClient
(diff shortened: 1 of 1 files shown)


josegonzalez:
Can we make this a conditional import for whatever version is configured?

airtoxin:
@josegonzalez Yes. Supports conditional importing was already PRed.
https://github.com/python-diamond/Diamond/pull/148

airtoxin:
If users install old version module and they only use old API, conditional branch will be needed, but I don't know how to do it.

josegonzalez:
So is #148 working for both versions?

airtoxin:
This is for new version module and influxdb version 0.8.
#148 is for old version module and influxdb version 0.8 or 0.9.

PR #189, opened 2015-05-23
Fix for-loop indentation error

patricklucas:
This causes diamond to crash on startup if exactly zero collectors are
found in the configured collectors directory.


mattrobenolt:
lol

PR #190, opened 2015-05-25
updated elasticsearch.py collector for alias stats

prune998:
This PR allows the user to change the Logstash regular expression in the config file.
While logstash always use the same regexp, some applications don't.

Default regexp is still the original logstash one...

### DIFF
--- a/src/collectors/elasticsearch/elasticsearch.py
+++ b/src/collectors/elasticsearch/elasticsearch.py
@@ -57,6 +57,11 @@ class ElasticSearchCollector(diamond.collector.Collector):
                 port = 9200
 
             self.instances[alias] = (host, int(port))
+        
+        if len(self.config['logstash_filter']):
+            re_logstash_filter = re.compile(self.config['logstash_filter'])
+        else:
+            re_logstash_filter = RE_LOGSTASH_INDEX
 
     def get_default_config_help(self):
         config_help = super(ElasticSearchCollector,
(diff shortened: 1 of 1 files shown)


src/collectors/elasticsearch/elasticsearch.py
```diff
@@ -57,6 +57,11 @@ def process_config(self):
                 port = 9200
 
             self.instances[alias] = (host, int(port))
+        
+        if len(self.config['logstash_filter']):
```
mattrobenolt: Won't this always be set since you're defining the default?

This could just be simplified to `re_logstash_filter = re.compile(self.config['logstash_filter'])`

src/collectors/elasticsearch/elasticsearch.py
```diff
@@ -90,6 +98,7 @@ def get_default_config(self):
             'path':           'elasticsearch',
             'stats':          ['jvm', 'thread_pool', 'indices'],
             'logstash_mode': False,
+            'logstash_filter': '^(.*)-\d\d\d\d\.\d\d\.\d\d$',
```
mattrobenolt: Either use the `RE_LOGSTASH_INDEX` here, or just bail on the constant entirely.

mattrobenolt:
Also, there's no way that this will work as-is. You're defining a new local var `re_logstash_filter` inside of `process_config` which is not accessible from within `_index_metrics` where it's attempting to be used.

PR #193, opened 2015-05-26
Fix collector discovery

patricklucas:
This fixes a bug I introduced while fixing a different bug in this logic.

Instead of iterating through all collector classes to find a match for
each configured collector, build a dict of cls_name -> cls and look up
collectors by name.


josegonzalez:
Seems reasonable.

bmhatfield:
I tested this and agree that it's reasonable and fixes the problem.

Thanks for the fix, @patricklucas !

PR #196, opened 2015-06-03
Changed to use the influxdb08 compatible module

justin8:
The latest version of influxdb python client does not work currently. This just sets it to use the v0.8x compatible module. influxdb 0.9x support should ideally be put in to a new module.


justin8:
Noticed someone else has a better pull request already in for this. just not merged yet.

PR #198, opened 2015-06-04
Reset signal handlers of collectors to default

benheilman:
Stops deletion attempt of the pidfile by collectors


josegonzalez:
Fixes #356

PR #203, opened 2015-06-09
Fix queues_ignored parameter

### LINKED ISSUE #202: [RabbitMQ] Parameter queues_ignored invalid.
All ini parameters are strings so when we set:
```
ignored_queues = foo
```
then in collector we would have list ['f', 'o', 'o'] instead of ['foo'].

Nihn:
Fix #202.

### DIFF
--- a/src/collectors/rabbitmq/rabbitmq.py
+++ b/src/collectors/rabbitmq/rabbitmq.py
@@ -106,7 +106,7 @@ class RabbitMQCollector(diamond.collector.Collector):
             'user': 'guest',
             'password': 'guest',
             'replace_dot': False,
-            'queues_ignored': [],
+            'queues_ignored': '',
             'cluster': False,
         })
         return config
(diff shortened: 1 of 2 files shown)


src/collectors/rabbitmq/rabbitmq.py
```diff
@@ -144,7 +144,7 @@ def collect(self):
         self.collect_health()
         matchers = []
         if self.config['queues_ignored']:
-                for reg in self.config['queues_ignored']:
+                for reg in self.config['queues_ignored'].split():
```
mattrobenolt: Do you want to use commas here instead of splitting on spaces? Do we use spaces elsewhere as a delimiter?

Nihn: Yes, spaces are used in [vhost] section like:
```
[vhost]
/ = queue1 queue2
```

mattrobenolt: That's gross, but ok. :)

mattrobenolt:
Thanks. :cake:

PR #206, opened 2015-06-15
Added multi-purpose docker collector

docwhat:
This captures data in a bit thorough way than memory_docker
as well as capturing the number of running and stopped containers.

Our use case is that we run integration tests using docker containers
and it is useful to see how they scale up and down during tests.


docwhat:
I could probably use some help doing the tests for the `collect()` method... I've tried mocking out the `docker.Client()` but it doesn't seem to work right.  My python is rusty and mocking didn't really exist when I did do python.

FYI: I don't understand why the python 2.7 build is failing... something about `cython`.

rtoma:
@docwhat Nice collector, especially like the image and dangling count!

That 2.7 travis error is strange, since a few days ago I passed the 2.7 checks with:
```
Collecting pyutmp (from -r .travis.requirements.txt (line 16))
  Downloading pyutmp-0.2.1.tar.gz
    file pyutmp.py (for module pyutmp) not found
```

So I triggered a re-check and now all's fine. :+1:

src/collectors/docker_collector/test/testdocker_collector.py
```diff
@@ -0,0 +1,165 @@
...
+            ])
+
+        m = mock_open(read_data=read_data)
+        with patch('__builtin__.open', m, create=True):
+            self.assertEqual(self.collector._default_memory_path(),
+                             '/mickeymouse/memory')
+
+        m.assert_called_once_with('/proc/mounts')
+
+    # @run_only_if_docker_client_is_available
```
mattrobenolt: Why were all of these tests commented out? :(

PR #207, opened 2015-06-16
Add oozie collector

johnseekins:
http://oozie.apache.org/

Collector gets data from the oozie instrumentation REST endpoint.


jaingaurav:
@johnseekins: What is the status of this PR? It doesn't seem like its passing its test cases.

johnseekins:
Not really sure how to fix this test error:
AttributeError: TestHTTPResponse instance has no attribute 'status'

It seems requests.get() fails in this test harness?

Issue #201, opened 2015-06-08
NetworkCollector not always collecting stats


justin8:
I'm using diamond on 5 machines currently, and only 2 of them actually report network stats.

I'm not sure how to debug the collector, but the only network configuration difference really is that the 3 that are not working are configured via networkmanager and the other two via netctl.

huoxy:
The default interfaces to collect are: eth, bond, em, p1p
you should add more within your situation

ref: https://github.com/BrightcoveOS/Diamond/wiki/collectors-NetworkCollector

justin8:
Ahh, I hadn't realized that. Thanks.

It might be worth updating it to support interface names under systemd considering almost all majors distros are using it now. (eno/ens/enp/enx/wlan/etc)

Full list of them are here: http://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/

If I get time I'll try and attach a merge request here sometime soon.

justin8:
Bump. This should be a quick and easy merge, it's mostly a config change.

shortdudey123:
Solved by https://github.com/python-diamond/Diamond/pull/208

PR #208, opened 2015-06-17
Added support for systemd interface names by default

justin8:
Added support for new interface permutations in systemd (http://www.freedesktop.org/wiki/Software/systemd/PredictableNetworkInterfaceNames/) from issue #201


Investocat:
:+1:

PR #210, opened 2015-06-19
Mesos collector

MichaelDoyle:


bmhatfield:
Looks good to me after the PEP8 issues are cleaned up.

MichaelDoyle:
@bmhatfield PEP-8 issues resolved :)

src/collectors/mesos/mesos.py
```diff
@@ -0,0 +1,73 @@
...
+    host = localhost
+    port = 5050
+```
+"""
+
+import diamond.collector
+import json
+import re
+import urllib2
+from pprint import pprint
```
mattrobenolt: This import is unused. plz remove.

src/collectors/mesos/mesos.py
```diff
@@ -0,0 +1,73 @@
...
+
+            response = urllib2.urlopen(url)
+            json_str = response.read()
+            return json.loads(json_str)
+        except (urllib2.HTTPError, ValueError), err:
+            self.log.error('Unable to read JSON response: %s' % err)
+            return {}
+
+    def clean_up(self, text):
+        text = re.sub('/', '.', text)
```
mattrobenolt: You don't need to use `re` here. This can be simplified to just:

```python
def clean_up(self, text):
  return text.replace('/', '.')
```

src/collectors/mesos/mesos.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def get_metrics(self):
+        try:
+            url = "http://%s:%s/%s" % (self.config['host'],
+                                       self.config['port'],
+                                       self.METRICS_PATH)
+
+            response = urllib2.urlopen(url)
+            json_str = response.read()
+            return json.loads(json_str)
```
mattrobenolt: This can be just:

```python
return json.load(urllib2.urlopen(url))
```

bmhatfield:
@mattrobenolt thanks for the feedback! We'll implement your suggestions on Monday :-)

MichaelDoyle:
@mattrobenolt @bmhatfield updated per feedback. thanks.

src/collectors/mesos/mesos.py
```diff
@@ -0,0 +1,68 @@
...
+
+    METRICS_PATH = "metrics/snapshot"
+
+    def get_default_config_help(self):
+        config_help = super(MesosCollector,
+                            self).get_default_config_help()
+        config_help.update({
+            'host': 'Hostname',
+            'port': 'Port',
+            'path': 'Metrics Path'
```
mattrobenolt: I don't think we should be overriding this here and providing a hard, absolute path value for all machines. Seems undesirable. What was your intent with this? From my understanding, this is going to lump everything under one toplevel key, `mesos` and we'd have no way to differentiate between machines.

src/collectors/mesos/mesos.py
```diff
@@ -0,0 +1,68 @@
...
+        })
+        return config_help
+
+    def get_default_config(self):
+        config = super(MesosCollector, self).get_default_config()
+        config.update({
+            'host': 'localhost',
+            'port': 5050,
```
mattrobenolt: Is there ever a case where the mesos master isn't running on 5050 and slave running on 5051? Would maybe be nice to just have a boolean to indicate if this is a master or a slave and choose the ports automatically, if we can assume that.

MichaelDoyle: 5050 and 5051 are the default values, however these ports are configurable by the user:

http://mesos.apache.org/documentation/latest/configuration/

What we could do, perhaps, is add a flag (is_master) as you've suggested, and pick the default port if not otherwise provided. I think we still need to allow the user to provide it, though, in the event that they have changed it. 

Sound good? Or does that complicate things too much? :)

src/collectors/mesos/mesos.py
MichaelDoyle: Just trying to line this up with what's in `get_default_config` down below. Looking at this, though, I don't think I need this here (in `get_default_config_help`). Just `get_default_config`.

Most (if not all?) collectors do provide a path in `get_default_config`. Examples:

* [CPU](https://github.com/python-diamond/Diamond/blob/2c0de81281a6750cf06fac760c081f89a088bca4/src/collectors/cpu/cpu.py#L56)
* [Disk](https://github.com/python-diamond/Diamond/blob/2c0de81281a6750cf06fac760c081f89a088bca4/src/collectors/diskusage/diskusage.py#L60)
* [Network](https://github.com/python-diamond/Diamond/blob/2c0de81281a6750cf06fac760c081f89a088bca4/src/collectors/network/network.py#L42)

The intent is for these to end up being something like `servers.hostname.mesos.master.value`.

mattrobenolt: Ah, sorry, I read over the code and misunderstood. I thought this would replace the entire `servers.<hostname>` part.

src/collectors/mesos/mesos.py
mattrobenolt: Nah, I think that complicates too much. I wasn't sure if you could set different ports or not, but since you can, there's not much value in saying `is_master` or anything. If anything, I'd just comment that or put it in the port's help text. Could just be `Port (default 5050, but needs to be set to 5051 for mesos-slave)`

MichaelDoyle:
Updated the help text and comments.

mattrobenolt:
Awesome, thanks a lot! :cake: :sparkles:

PR #212, opened 2015-06-26
fix batch_max_interval handling in signalfx handler

baris:
Fixes TypeError handling batch_max_interval configuration option.

### DIFF
--- a/src/diamond/handler/signalfx.py
+++ b/src/diamond/handler/signalfx.py
@@ -35,14 +35,14 @@ class SignalfxHandler(Handler):
         self.batch_size = int(self.config['batch'])
         self.url = self.config['url']
         self.auth_token = self.config['auth_token']
-        self.batch_max_interval = self.config['batch_max_interval']
+        self.batch_max_interval = int(self.config['batch_max_interval'])
         self.resetBatchTimeout()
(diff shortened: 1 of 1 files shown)


src/diamond/handler/signalfx.py
```diff
@@ -35,14 +35,14 @@ def __init__(self, config=None):
...
         self.auth_token = self.config['auth_token']
-        self.batch_max_interval = self.config['batch_max_interval']
+        self.batch_max_interval = int(self.config['batch_max_interval'])
         self.resetBatchTimeout()
         if self.auth_token == "":
             logging.error("Failed to load Signalfx module")
             return
 
     def resetBatchTimeout(self):
-        self.batch_max_timestamp = int(time.time() + self.batch_max_interval)
```
jaingaurav: Since this is going to be compared with time.time() which returns a float, I think the cast should be to a float instead of an int to avoid loosing precision. In that case, you may want to change `self.batch_max_interval` to be a float instead of an int as well.

jaingaurav:
Thanks, this looks great. Please fix the cast issue noted and I can merge this.

shortdudey123:
Still needs casting issue fixed.

PR #213, opened 2015-07-06
Should not specify `--foreground` when running as a daemon

cannium:


mzupan:
not sure what this is doing.. upstart wants the service to be in the foreground. 

This might be launching the init.d script.. Generally how it works is the init.d script will call the upstart script. I've not touched RHEL7 yet though.

cannium:
`upstart wants the service to be in the foreground` -- could you give me some pointers?

mzupan:
pointers with what? things like supervisord/upstart want the running services to be in the foreground so they can monitor to tell if they are running or crashed so they can try to restart them.

cannium:
OK it is a bad patch. But diamond started with `--foreground` doesn't write a pid file so `service diamond status` returns wrong answer. I suppose that to be a bug.

mzupan:
yes that might be a bug.. i'd have to load a rhel7 VM to test but we can make upstart write a pid. You can open an issue on that for tracking purposes.

cannium:
Opened an issue: #215

PR #216, opened 2015-07-06
Update AWS CloudWatch handler to use the proper dimension name

tommysdk:
The CloudWatch handler sets the instance dimension name to "InstanceID", AWS however specify their default metrics with the name "InstanceId", which is also is the expected casing used elsewhere as well withing AWS. The Diamond CloudWatch handler should obey this naming convention to allow metrics to be published within the same dimension as e.g. the default AWS EC2 metrics (CPUUtilization, DiskReadBytes etc).

It would make it more coherent when e.g. configuring CloudWatch alarms through Cloudformation when the same dimension name "InstanceId" could be used for all metrics, instead of needing to know that Diamond pushed metrics use another case than the AWS ones.

### DIFF
--- a/src/diamond/handler/cloudwatch.py
+++ b/src/diamond/handler/cloudwatch.py
@@ -2,7 +2,7 @@
 """
 Output the collected values to AWS CloudWatch
 
-Automatically adds the InstanceID Dimension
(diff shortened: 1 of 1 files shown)


jaingaurav:
lgtm

PR #223, opened 2015-07-16
XFS stats collector

omame:


src/collectors/xfs/test/testxfs.py
```diff
@@ -0,0 +1,201 @@
...
+        self.collector = XFSCollector(config, None)
+
+    def test_import(self):
+        self.assertTrue(XFSCollector)
+
+    @patch('__builtin__.open')
+    @patch('os.access', Mock(return_value=True))
+    @patch.object(Collector, 'publish')
+    def test_should_open_proc_stat(self, publish_mock, open_mock):
+        open_mock.return_value = StringIO('')
```
jaingaurav: Is this the only place where StringIO is used? Isn't there a simpler way?

jaingaurav:
Only really had a very minor comment. But this looks great. I'll merge it in.

PR #224, opened 2015-07-17
Enabling signalfx handler to use the signalfx python library

ramjothikumar:
- The signalfx python library has support for Protocol Buffers. To take advantage of this and other enhancements added to the python library, this patch enables the signalfx handler to use the python library.
  This change adds a dependency - signalfx
- Changing batch_max_timestamp type to float to be more accurate during the should_flush calculation.


src/diamond/handler/signalfx_handler.py
```diff
@@ -5,44 +5,50 @@
...
         self.batch_size = int(self.config['batch'])
-        self.url = self.config['url']
-        self.auth_token = self.config['auth_token']
-        self.batch_max_interval = self.config['batch_max_interval']
+        self.batch_max_interval = float(self.config['batch_max_interval'])
         self.resetBatchTimeout()
-        if self.auth_token == "":
-            logging.error("Failed to load Signalfx module")
-            return
+        self.signalfx = signalfx.SignalFx(auth_token, ingest_endpoint=url,
```
baris: Looks like this uses the default 1s timeout. Making api post timeout a configuration option for the handler would be very useful.

ramjothikumar: Agree, will add that.

nathanleiby:
Would love to see this merged.

nathanleiby:
I also just opened another Signalfx PR, fwiw: https://github.com/python-diamond/Diamond/pull/236

ramjothikumar:
Whats the best way for me to expedite this getting merged.

src/diamond/handler/signalfx_handler.py
```diff
@@ -5,44 +5,57 @@
...
+        auth_token = self.config['auth_token']
+        url = self.config['url']
         self.batch_size = int(self.config['batch'])
-        self.url = self.config['url']
-        self.auth_token = self.config['auth_token']
-        self.batch_max_interval = self.config['batch_max_interval']
+        self.batch_max_interval = float(self.config['batch_max_interval'])
+        self.timeout = self.config['timeout']
         self.resetBatchTimeout()
-        if self.auth_token == "":
```
shortdudey123, line 40: Don't we want an error still logged about no api key?  Or does the signalfx library handle that?

PR #226, opened 2015-07-17
bugfix: type error on values = []

h00dy:
Fix in HTTPJSONCollector
When one of a key values is a list, the collector fails silently with:
TypeError: int() argument must be a string or a number, not 'list'


jaingaurav:
Looks good. Could you please add a test case? Probably would be good to test catching other types like string values e.g. "blah"

Note that It seems like the `testhttpjson.py` needs to migrate from using the pickled output and `self.getPickledResults` to using `self.assertPublishedMany` as with many other test cases.

PR #228, opened 2015-07-18
add multicollector support based on whitespace and dot split

jriguera:
Currently diamond is able to create multiple instances per collector class by loading config files with the format  `CollectorClassName instance.conf`, in my opinion I think is better if it can support also dots as separator to make it easy for shell scripts, provisioning tools, and also for some people (like me :-). What do do think?

If the patch is accepted I will create a record in the FAQ.md explaining those options to get them documented.

### DIFF
--- a/src/diamond/server.py
+++ b/src/diamond/server.py
@@ -161,10 +161,11 @@ class Server(object):
                 )
 
                 for process_name in running_collectors - running_processes:
-                    # To handle running multiple collectors concurrently, we
-                    # split on white space and use the first word as the
-                    # collector name to spin
+                    # To handle running multiple collectors concurrently, fist
+                    # we split on white space, then on dot and use the first
+                    # word as the collector name to spin.
                     collector_name = process_name.split()[0]
+                    collector_name = collector_name.split('.', 1)[0]
 
                     if 'Collector' not in collector_name:
                         continue


mattrobenolt:
I guess I'm curious what cases make it easier to use dots instead of spaces?

mattrobenolt:
In your example, what would it change to?

`CollectorClassName.instance.conf` ?

jriguera:
yep, in my example it will be:
`CollectorClassName.instance1.conf`
`CollectorClassName.instance2.conf`

It is about adding more flexibility, it does not break compatibility, people can still use spaces.
For example, when there are some instances with white spaces and using the tab 
bash completion:
`CollectorClassName.conf`
`CollectorClassName instance1.conf`
`CollectorClassName instance2.conf`
by using dot, make it easier typing `[.][tab]` than `[\][ ][tab]` to autocomplete.

The same happens if one wants to create a quick script and forgets
the `"` with the variables ... like this:
`$ for i in CollectorClassName*.conf; do sed -i 's/^enabled.*/enabled = False/g'  $i; done`

mattrobenolt:
Ok, yeah, I get it. I just firmly disagree. I don't think that justification is enough to warrant a weird configuration option that we'll forget about when it causes a bug at some point in the future just to save a keystroke for tab completion or a bash script.

Thanks though for the suggestion.

PR #233, opened 2015-07-27
Add systemd collector

gmlexx:
Send 1 if configured systemd service status is running and 0 otherwise

### DIFF
--- a/src/collectors/elasticsearch/elasticsearch.py
+++ b/src/collectors/elasticsearch/elasticsearch.py
@@ -152,7 +152,7 @@ class ElasticSearchCollector(diamond.collector.Collector):
         # publish all 'total' and 'time_in_millis' stats
         self._copy_two_level(
             metrics, prefix, index,
-            lambda key: key.endswith('total') or key.endswith('time_in_millis'))
+            lambda key: key.endswith('total') or key.endswith('time_in_millis') or key.endswith('in_bytes') or key.endswith('count'))
 
     def _add_metric(self, metrics, metric_path, data, data_path):
         """If the path specified by data_path (a list) exists in data,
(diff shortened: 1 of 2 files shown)


src/collectors/systemd/systemd-collector.py
```diff
@@ -0,0 +1,53 @@
...
+            manager = Manager()
+
+            for service in [s.strip() for s in self.config['services']]:
+
+                if not service:
+                    continue
+
+                try:
+                    unit = manager.get_unit('%s.service' % service)
+                    self.publish("%s.running" % service, 1
```
bmhatfield: Are there more statistics about systemd services we can send here, beyond just whether or not a service is enabled?

gmlexx: There are many other properties that we can access (look at http://www.freedesktop.org/wiki/Software/systemd/dbus/ Unit Objects properties), but there are not so useful as is service actually running or not.

src/collectors/elasticsearch/elasticsearch.py
```diff
@@ -152,7 +152,7 @@ def _index_metrics(self, metrics, prefix, index):
         # publish all 'total' and 'time_in_millis' stats
         self._copy_two_level(
             metrics, prefix, index,
-            lambda key: key.endswith('total') or key.endswith('time_in_millis'))
+            lambda key: key.endswith('total') or key.endswith('time_in_millis') or key.endswith('in_bytes') or key.endswith('count'))
```
eMerzh: Is it really meant to be in the same PR ? i don't see the relation....?!

gmlexx: Surely not. PR was from master. One non-related commit has been added.

PR #236, opened 2015-07-30
signalfx handler: allow filtering sent metrics

nathanleiby:
This allows the Signalfx Handler to filter which fields are sent as metrics. It's taken from the Librato handler, which has the same feature: https://github.com/python-diamond/Diamond/blob/2c0de81281a6750cf06fac760c081f89a088bca4/src/diamond/handler/libratohandler.py#L26


nathanleiby:
Is there a documented PR process one should follow? Thanks!

PR #237, opened 2015-07-30
mongodb: collect Mongo replSet status

nathanleiby:
Fetches Mongodb replica set stats by calling [replSetGetStatus](http://docs.mongodb.org/manual/reference/command/replSetGetStatus/). Also publishes keys `healthy_nodes` and `total_nodes` as a summary of the per-node data.

This can be enabled by setting "replica" in the config.

### DIFF
--- a/src/collectors/mongodb/mongodb.py
+++ b/src/collectors/mongodb/mongodb.py
@@ -4,6 +4,10 @@
 Collects all number values from the db.serverStatus() command, other
 values are ignored.
 
+**Note:** this collector expects pymongo 2.4 and onward. See the pymongo
+changelog for more details:
+http://api.mongodb.org/python/current/changelog.html#changes-in-version-2-4
+
 #### Dependencies
 
  * pymongo
--- a/src/collectors/mongodb/mongodb.py
+++ b/src/collectors/mongodb/mongodb.py
@@ -59,13 +59,17 @@ class MongoDBCollector(diamond.collector.Collector):
             'of collections. This is applied after collections are ignored via'
             ' ignore_collections Sampling uses crc32 so it is consistent across'
             ' replicas. Value between 0 and 1. Default is 1',
-            'network_timeout': 'Timeout for mongodb connection (in seconds).'
-                               ' There is no timeout by default.',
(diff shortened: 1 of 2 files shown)


src/collectors/mongodb/mongodb.py
```diff
@@ -147,16 +152,15 @@ def collect(self):
                     self.config['ssl'] = str_to_bool(self.config['ssl'])
 
                 if ReadPreference is None:
-                    conn = pymongo.Connection(
+                    conn = pymongo.MongoClient(
```
bmhatfield: What precipitated this change? Will users need to update pymongo to be compatible?

andrewhuang: This change allowed us to use pymongo 3.x, which [changed the client class constructor when moving from 2.x to 3.0](http://api.mongodb.org/python/current/changelog.html#mongoclient-changes). There should be no need to upgrade the pymongo client if they are using a relatively up to date version of 2.x (Connection has been depreciated since 2.5 at least). However, users should note that the network_timeout value is now in milliseconds when configuring this collector.

nathanleiby:
@bmhatfield thoughts on merging this change or additional compatibility we need to figure out first? Thanks!

bmhatfield:
Sorry for the delay. First of all, thanks for this change.

After reading your commentary, I think there's unfortunately a risk of breaking existing installations with this change. Can you add documentation about compatibility (Specific version ranges if you can find them) to the header comments for this collector? I don't think we should be stuck in the past, but looking at my 12.04 mongo box, I see that for whatever reason (low effort invested here), I ended up with `pymongo (2.1)`. If the deprecation happened in 2.5, I think there's a chance this PR will break my existing collectors. I can probably upgrade without issue, but others may not be able to, so there should hopefully at least be information about why they stopped working on upgrade in the comments.

andrewhuang:
@bmhatfield I've added a note accordingly. Is there anything else we should do to push this along?

andrewhuang:
@bmhatfield: tests are green after rerunning them. Let me know if there's anything else to do. :smile:

bmhatfield:
Thanks!

PR #240, opened 2015-08-03
adapt to changed psutil functions interface

grobian:
psutil functions were renamed, apply those changes so Diamond works again using latest psutil.


josegonzalez:
This breaks existing installations, so maybe you want to check if `hasattr` before using the new versions, else use the old?

grobian:
Is it really worth doing so?  Can't we require a newer version?  On Linux Diamond still uses its own parsing, so never uses psutil, isn't it?

jaingaurav:
@grobian: I think we need diamond to be compatible with both the new and the old one since it it possible that a user may not be able to upgrade due to some other dependancies on the old pustules.

shortdudey123:
Can you address the above concern and squash your commits?

grobian:
No, I've moved on and no longer have this, feel free to use parts that you want though.

Issue #241, opened 2015-08-04
Log file permissions are world-writable


cjdaniel:
Running Diamond on Debian Wheezy built with the included scripts, log files in /var/log/diamond end up with permissions that are too permissive. Example, after a few days of log rotation:

```
$ ls -l /var/log/diamond
total 19752
-rw-rw-rw- 1 root root 2207020 Aug  4 18:52 archive.log
-rw-rw-rw- 1 root root  388605 Jul 28 23:59 archive.log.2015-07-28
-rw-rw-rw- 1 root root 2810123 Jul 29 23:59 archive.log.2015-07-29
-rw-rw-rw- 1 root root 2835343 Jul 30 23:59 archive.log.2015-07-30
-rw-rw-rw- 1 root root 3008676 Jul 31 23:59 archive.log.2015-07-31
-rw-rw-rw- 1 root root 2809072 Aug  1 23:59 archive.log.2015-08-01
-rw-rw-rw- 1 root root 2809079 Aug  2 23:59 archive.log.2015-08-02
-rw-rw-rw- 1 root root 2809082 Aug  3 23:59 archive.log.2015-08-03
-rw-rw-rw- 1 root root    1969 Aug  4 18:05 diamond.log
-rw-rw-rw- 1 root root      59 Jul 29 21:44 diamond.log.2015-07-28
-rw-rw-rw- 1 root root      59 Aug  4 17:27 diamond.log.2015-07-29
-rw-rw-rw- 1 root root   92919 Jul 30 23:59 diamond.log.2015-07-30
-rw-rw-rw- 1 root root   92919 Jul 31 23:59 diamond.log.2015-07-31
-rw-rw-rw- 1 root root   92751 Aug  1 23:59 diamond.log.2015-08-01
-rw-rw-rw- 1 root root   92751 Aug  2 23:58 diamond.log.2015-08-02
-rw-rw-rw- 1 root root   92751 Aug  3 23:58 diamond.log.2015-08-03
```

As you can see, these files are world-writable. The logging handler being used does not have any options for setting permissions, so it seems the umask set for diamond itself is effective here.

See this line: https://github.com/python-diamond/Diamond/blob/master/bin/diamond#L239

Changing this to umask(0o022) fixes the problem, and doesn't seem to cause any other problems, at first glance.

jaingaurav:
PR #242 was merged

PR #242, opened 2015-08-04
Fix for #241 (Log file permissions world-writable)

cjdaniel:
Sets umask to 022 instead of 000, to avoid creating new files as world-writable.


jaingaurav:
lgtm thanks

Issue #227, opened 2015-07-17
Hadoop collector config as list


ethan-eb:
Ran into this today ... In my collectors/HadoopCollector.conf I had 

```
metrics = /mnt/log/hadoop-metrics/*.out
```

but was not getting any hadoop metrics out. Dug around, manually ran the collector, and guessed that the setting was not being treated as a list in https://github.com/python-diamond/Diamond/blob/master/src/collectors/hadoop/hadoop.py#L46 I tried setting my metrics config to 

```
metrics = /mnt/log/hadoop-metrics/*.out,
```

and it worked out fine.

I can put together a PR - but prob not today, just wanted to register the issue before I forget it!

PR #243, opened 2015-08-05
Allow single hadoop metrics file python-diamond/Diamond#227

ethan-eb:


ethan-eb:
Travis check doesn't look related to my change - it's trying to pip install travis requirements `pip install -r .travis.requirements.txt`

src/collectors/hadoop/hadoop.py
```diff
@@ -43,7 +43,10 @@ def get_default_config(self):
         return config
 
     def collect(self):
-        for pattern in self.config['metrics']:
+        conf_metrics = self.config['metrics']
+        if type(conf_metrics) == str:
```
jaingaurav: So we generally follow this convention for converting strings to lists:
```python
if isinstance(conf_metrics, basestring):
    conf_metrics = [conf_metrics]
```

jaingaurav:
Minor comment in order to support unicode strings. Otherwise looks good. Not sure what's up with the travis build.

ethan-eb:
Ok makes sense, updated!

ebpitts:
Any reason this can't be merged?

Issue #234, opened 2015-07-27
HadoopCollector can't use white/black lists


ethan-eb:
I found that my whitelist configuration was being ignored for HadoopCollector, in reading the code I see that it uses self.publish_metric() instead of self.publish() at the end of it's collect() routine. In other collectors the publish method is called, which filters the white/black lists and then calls publish_metrics. 

It should be possible to switch it to publish() the only complication is the way HadoopCollector is building the metric path.

ethan-eb:
One other thing: comparing whitelist / blacklist against the name and then setting the path afterwards seems incorrect. Wouldn't you expect to be able to pattern match against the full path?

It is too limited in the particular case of Hadoop type metrics. You get (for example) a **name** such as *metricssystem.MetricsSystem* ... the path then gets a variety of metrics appended to that *metricssystem.MetricsSystem.Context* *metricssystem.MetricsSystem.NumActiveSources* *metricssystem.MetricsSystem.PublishNumOps* etc, etc.

In any case, changing the existing behavior seems like it could cause unexpected changes with existing configurations in the wild. 

I'm going to try adding a path_filter method, and then provide a way for a collector to choose the path filter if wanted from it's publish method. Will send a pull request...

shortdudey123:
Merged #245.

PR #244, opened 2015-08-05
Filter hadoop collector for whitelist/blacklist 

ethan-eb:
python-diamond/Diamond#234

This change adds a new path_allowed method that checks white/blacklists against full path instead of just the "name." This is desirable in hadoop because there are path variations for a given name that might need to be handled differently.

The new method is called from the hadoop collector, so it doesn't affect existing white/blacklist behaviors for other collectors.

In the test I added a 'metrics_blacklist' config, confirmed that broke the test, and then adjusted expected.pkl to correct the test.


src/collectors/hadoop/hadoop.py
```diff
@@ -101,6 +101,10 @@ def collect_from(self, filename):
                             data['name'],
                             metric, ]))
 
+                    if not self.path_allowed(path):
+                        # whitelist / blacklist
+                        continue
+
                     value = float(metrics[metric])
 
                     self.publish_metric(Metric(path,
```
jaingaurav: Weren't you going to switch to the publish method?

ethan-eb: So I was trying to keep the issues and pull requests isolated, but there is overlap and so it is somewhat confusing. In my PR for multi black/white lists #245 I did move it to the publish method. Also renamed it metric_allowed.

I can combine these two things (path vs name & single vs multiple white/black lists) into one PR. Whatever you think makes more sense.

I'm going to rebase my existing PRs with the latest now...

ebpitts:
If #245 is merged this can be closed, since it's a subset of that PR.

shortdudey123:
Subset of https://github.com/python-diamond/Diamond/pull/245

Issue #230, opened 2015-07-20
Multiple whitelist / blacklist config settings


ethan-eb:
The hadoop collector produces a lot of metrics (I'm going to end up using the cassandra collector as well which is a similar case), and I'd like to filter those with metrics_whitelist. However, that setting only accepts a single regex. I don't think that it will be manageable going forward, there will be too many patterns to combine into a single monstrous regex.

I'm interested in adjusting that so that you can provide multiple regex patterns. For both whitelist and blacklist it could stop as soon as it finds a positive match (and proceed or stop as appropriate) - or continue to the end of the list if no match is found and then proceed or stop.

I am happy to put together a pull request for this, but first wanted to ask if it's something that sounds interesting...

ethan-eb:
One other thing: comparing whitelist / blacklist against the *name* and then setting the path afterwards seems incorrect. Wouldn't you expect to be able to pattern match against the full path?

In any case, changing that behavior seems like it could cause unexpected changes with existing configurations in the wild. I'll submit a pull request, but I haven't decided if I should change that.

Feedback appreciated.

shortdudey123:
Merged #245.

PR #245, opened 2015-08-05
Multi blackwhitelist

ethan-eb:
I adjusted collector.py code to handle multiple whitelists or
blacklists, and used the HadoopCollector to test the changes.

The collector.process_config now ensures that
collector.config['metrics_whitelist'](and blacklist) is either a list
of regex objects, or an empty list.

collector.metric_allowed is based on work from a previous commit and is
now generalized enough use either with path or name filtering. It now
checks white/black lists in a loop. The collector.publish uses that same
method now to avoid code duplication.

For testing I first confirmed that I broke tests by making
'metrics_blacklist' in testhadoop.py a tuple (because the code expected
a string). I then made my changes and confirmed the test was still
broken because the newly blacklisted path was not in the list of
expected metrics. I then edited expected.pkl to remove that metric, and
confirmed tests passed once again.


ethan-eb:
So this PR includes the fix for python-diamond/Diamond#230 and python-diamond/Diamond#234 because I wanted to build on the changes in the #234 fix. If you all don't want this multi white/blacklists stuff, there is a separate PR which is *only* the #234 fix : https://github.com/python-diamond/Diamond/pull/244

nathanleiby:
:100: for making this generic in collectors. 

I faced a similar issue in https://github.com/python-diamond/Diamond/pull/236 and think a generic whitelist/blacklist for handlers would also be useful!

ebpitts:
The test failures reported by Travis look spurious.

https://travis-ci.org/python-diamond/Diamond/jobs/94716665 = pep8 violations in collectors not touched by this PR

https://travis-ci.org/python-diamond/Diamond/jobs/94716664 = pep8 violations plus some strange import and test function errors, maybe related to python 2.6?

Is this mergeable?

shortdudey123:
@josegonzalez or @jaingaurav can you guys review this as well since its a change to a core config options?

PR #246, opened 2015-08-07
DiskSpace : Add nfs4 as default FS to monitor

eMerzh:
On debian based distributions nfs is often called nfs4 , it might be a good idea to add in the default list ...

### DIFF
--- a/src/collectors/diskspace/diskspace.py
+++ b/src/collectors/diskspace/diskspace.py
@@ -51,8 +51,8 @@ class DiskSpaceCollector(diamond.collector.Collector):
         config.update({
             'path': 'diskspace',
             # filesystems to examine
-            'filesystems': 'ext2, ext3, ext4, xfs, glusterfs, nfs, ntfs, hfs,'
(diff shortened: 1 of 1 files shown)


rtoma:
Hi,
Nfs4's fine with me, but ntfs and hfs seems too special to be part of the default.
If anyone needs these, they can easilly add them in the collector config file.
Maybe I am biased. Anyone?

mattrobenolt:
I'd prefer to not change backwards compatibility. Can we just add `nfs4` and leave the others alone?

bmhatfield:
I agree with @rtoma and @mattrobenolt.

mattrobenolt:
I think @rtoma misread the diff. `ntfs` and `hfs` aren't being added. They're just shifting over to make place for `nfs4`. :)

So this patch is only adding `nfs4`, which lgtm. :+1:

bmhatfield:
Oh no :-( I misread it too. Sorry folks.

PR #249, opened 2015-08-13
Fixed cephstats to handle missing client io output. 

wyllys66:


src/collectors/cephstats/cephstats.py
```diff
@@ -8,12 +8,38 @@
...
+                iops = number.search(ceph_stats, wr.end())
+        ret = {}
+        if rd:
+            ret['rd'] = rd.group()
+        if wr:
+            ret['wr'] = wr.group()
+        if iops:
+            ret['iops'] = iops.group()
+        return ret
+    except:
```
mattrobenolt: What exceptions may be raised here? Please don't use a bare `except:`.

src/collectors/cephstats/cephstats.py
```diff
@@ -8,12 +8,38 @@
...
 sys.path.insert(0, os.path.join(os.path.dirname(os.path.dirname(__file__)),
                                 'ceph'))
 from ceph import CephCollector
 
 
+# This is external to the CephCollector so it can be tested
+# separately.
+def process_ceph_status(output):
+    pattern = re.compile(r'\bclient io .*')
+    res = pattern.search(output)
```
mattrobenolt: Couldn't we simply check `if 'client io ' not in output:` ?

src/collectors/cephstats/cephstats.py
wyllys66: RE could throw something, though probably not.  I think its safe to just delete the 'try/except' statements here but I wasnt sure.

src/collectors/cephstats/cephstats.py
wyllys66: Yes but it doesn't really save much.

src/collectors/cephstats/cephstats.py
mattrobenolt: Only thing that could possibly happen, but it seems that you're guarding against it, would be `re.search` returning `None`, then checking `.group()` which would raise an `AttributeError`.

But that case seems covered.

src/collectors/cephstats/cephstats.py
mattrobenolt: orly

```
$ python -m timeit -s 'import re' 'pattern = re.compile(r"\bclient io .*"); pattern.search("foo client io bar")'
1000000 loops, best of 3: 1.88 usec per loop
$ python -m timeit '"client io " in "foo client io bar"'
10000000 loops, best of 3: 0.0757 usec per loop
```

It's considerably faster to not use a regular expression. :)

wyllys66: ok, ill fix this and the other issue and resubmit

src/collectors/cephstats/cephstats.py
```diff
@@ -8,12 +8,38 @@
...
+        number = re.compile(r'\d+')
+        rd = number.search(ceph_stats)
+        if rd:
+            wr = number.search(ceph_stats, rd.end())
+            if wr:
+                iops = number.search(ceph_stats, wr.end())
+        ret = {}
+        if rd:
+            ret['rd'] = rd.group()
+        if wr:
```
mattrobenolt: This could raise a `NameError`.

src/collectors/cephstats/cephstats.py
```diff
@@ -8,12 +8,38 @@
...
+            ret['wr'] = wr.group()
+        if iops:
```
mattrobenolt: This could raise a `NameError` too.

src/collectors/cephstats/cephstats.py
```diff
@@ -8,12 +8,38 @@
...
+def process_ceph_status(output):
+    pattern = re.compile(r'\bclient io .*')
+    res = pattern.search(output)
+    if not res:
+        return {}
+    try:
+        ceph_stats = res.group()
+        number = re.compile(r'\d+')
+        rd = number.search(ceph_stats)
+        if rd:
```
mattrobenolt: If this were falsey, you'd get the `NameErrors` below.

wyllys66: how so?  If 'rd' is False or None, the others will remain undefined (i.e. None) and the "if" checks later will not pass.

src/collectors/cephstats/cephstats.py
```diff
@@ -8,12 +8,38 @@
...
+# This is external to the CephCollector so it can be tested
+# separately.
+def process_ceph_status(output):
+    pattern = re.compile(r'\bclient io .*')
+    res = pattern.search(output)
+    if not res:
+        return {}
+    try:
+        ceph_stats = res.group()
+        number = re.compile(r'\d+')
```
mattrobenolt: This `re.compile` call should be hoisted up to the module level.

src/collectors/cephstats/cephstats.py
mattrobenolt: That's now how Python works. :)

You're only defining those variables inside this if statement. So if this code branch is never hit because `rd` is falsey, they vars never get defined, resulting in a `NameError` when trying to check them later.

mattrobenolt:
```
>>> rd = False
>>> if rd:
...   wr = True
...
>>> if wr:
...   print 'hello'
...
  File "<stdin>", line 1, in <module>
NameError: name 'wr' is not defined
```

wyllys66: got it. thanks.

wyllys66:
Updated with suggestions from comments.

src/collectors/cephstats/cephstats.py
```diff
@@ -8,11 +8,40 @@
...
+
+# This is external to the CephCollector so it can be tested
+# separately.
+def process_ceph_status(output):
+    if 'client io' not in output:
+        return {}
+    res = patternchk.search(output)
+    if not res:
+        return {}
+    ceph_stats = res.group()
```
mattrobenolt: meh, I guess if you need to extract the group now, then you can skip the `if 'client io' not in output:` check. I feel like this wasn't the case before.

src/collectors/cephstats/cephstats.py
```diff
@@ -8,11 +8,40 @@
...
+    if not ceph_stats:
+        return {}
+    rd = wr = iops = None
+    rd = numberchk.search(ceph_stats)
+    if rd:
```
mattrobenolt: I'd check these against an explicit `if rd is None:`, `if wr is not None:` and `if iops is not None:`.

mattrobenolt:
Other than those two comments, this looks good to me.

wyllys66:
You want me to update with changes or is this good enough?

mattrobenolt:
Yes, please. :)

wyllys66:
Updated with explicit checks.

mattrobenolt:
lgtm, thanks! :cake: :sparkles:

PR #254, opened 2015-08-21
Ignore per project pyenv settings

Investocat:
When using pyenv, it's useful to set the python version for the project itself. This just ignores the dot file that is stored in the root of the project directory.


jaingaurav:
Seems reasonable, thanks!

PR #255, opened 2015-08-21
Fix collectors to support sub-second collection intervals

Investocat:
When testing collectors, some times it best to collect at a very rapid speed. The collector module attempts to cast the interval as an int rather then as a float and throws an exception. This fixes that


jaingaurav:
lgtm

PR #256, opened 2015-08-21
Prevent Diamond from hemorrhaging memory

### LINKED ISSUE #84: Diamond hemorrhages memory when Graphite server is inaccessible
I recently ran into a situation on one of my hosts where Diamond 4.0 series was consuming over 3 GB of memory after being unable to connect to Graphite for a couple of hours -- RES in htop appears to grow by 2-3 MB per minute. I'm not sure if this is an issue with unbounded write buffer growth or leaked objects in connection handling, but it presents a serious threat to system stability.

Investocat:
Diamond uses a unbounded queue to send metrics to the handler process When a handler stalls due to a connection attempt or similar, this causes the queue process to just grow until we run out of memory. This change does a few things.

1. Brounds the queue to 16k unprocessed metrics before dropping them
2. Removes a unneeded array and just sends the metrics directly down the queue directly rather then wrapping them in a list
3. Uses a None object as a flush indicator
4. Moves to __slots__ for the Metric object. This reduces the memory of the Metrics significantly
5. Adds methods to the Metric object to allow pickling/unpickling if ever needed

In my tests, this reduces the memory cost of 2048 Metric objects from ~35M to ~2M

This should fix #84 and even reduce the latency for the metrics being published.

### DIFF
--- a/src/diamond/metric.py
+++ b/src/diamond/metric.py
@@ -7,8 +7,14 @@ from error import DiamondException
 
 
 class Metric(object):
-
-    _METRIC_TYPES = ['COUNTER', 'GAUGE']
+    # This saves a significant amount of memory per object. This only matters
+    # due to the queue system that moves objects between processes and can end
+    # up storing a large number of objects in the queue waiting for the
+    # handlers to flush.
+    __slots__ = [
+        'path', 'value', 'raw_value', 'timestamp', 'precision',
+        'host', 'metric_type', 'ttl'
+        ]
 
     def __init__(self, path, value, raw_value=None, timestamp=None, precision=0,
                  host=None, metric_type='COUNTER', ttl=None):
@@ -24,7 +30,7 @@ class Metric(object):
         """
 
         # Validate the path, value and metric_type submitted
-        if (None in [path, value] or metric_type not in self._METRIC_TYPES):
+        if (None in [path, value] or metric_type not in ['COUNTER', 'GAUGE']):
             raise DiamondException(("Invalid parameter when creating new "
                                     "Metric with path: %r value: %r "
                                     "metric_type: %r")
@@ -78,6 +84,17 @@ class Metric(object):
 
         # Return formated string
         return fstring % (self.path, self.value, self.timestamp)
+
+    def __getstate__(self):
+        return dict(
+            (slot, getattr(self, slot))
+            for slot in self.__slots__
+            if hasattr(self, slot)
+        )
+
+    def __setstate__(self, state):
+        for slot, value in state.items():
+            setattr(self, slot, value)
 
     @classmethod
     def parse(cls, string):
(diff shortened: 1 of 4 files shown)


src/diamond/metric.py
```diff
@@ -7,8 +7,14 @@
 
 
 class Metric(object):
-
-    _METRIC_TYPES = ['COUNTER', 'GAUGE']
+    # This saves a significant amount of memory per object. This only matters
+    # due to the queue system that moves objects between processes and can end
+    # up storing a large number of objects in the queue waiting for the
+    # handlers to flush.
+    __slots__ = [
```
mattrobenolt: +1 for `__slots__`. Underutilized python feature. :)

src/diamond/metric.py
```diff
@@ -24,7 +30,7 @@ def __init__(self, path, value, raw_value=None, timestamp=None, precision=0,
         """
 
         # Validate the path, value and metric_type submitted
-        if (None in [path, value] or metric_type not in self._METRIC_TYPES):
+        if (None in [path, value] or metric_type not in ['COUNTER', 'GAUGE']):
```
mattrobenolt: Should use tuple here since we're changing this.

Investocat: Doh! Yeah, should have done that. Just committed that change.

josegonzalez: Why tuple vs list?

mattrobenolt:
```
>>> dis.dis(lambda: ['a','b'])
  1           0 LOAD_CONST               1 ('a')
              3 LOAD_CONST               2 ('b')
              6 BUILD_LIST               2
              9 RETURN_VALUE
>>> dis.dis(lambda: ('a','b'))
  1           0 LOAD_CONST               3 (('a', 'b'))
              3 RETURN_VALUE
```

Defining a tuple is an O(1), but a list is O(n). Minor in this case, just good hygiene.

src/diamond/server.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, configfile):
         self.manager = multiprocessing.Manager()
         if setproctitle:
             setproctitle(oldproctitle)
-        self.metric_queue = self.manager.Queue()
+        self.metric_queue = self.manager.Queue(maxsize=16384)
```
jaingaurav: Could you move this value to be a configuration parameter?

ltsampros:
This is indeed a very desirable.

zzzuzik:
+1

bmfurtado:
Could the maximum size for the unprocessed metrics queue be made configurable? I think choosing an arbitrary number of 16k is a bit odd :)

Other than that, :+1: !

jaingaurav:
@Investocat Any progress on this? We just need a config for this please.

jraby:
Nice memory usage improvement with `__slots__`,  I had overlooked that python feature until now!

See #330 for the same PR with the configurable queue size.

jaingaurav:
Thanks @jraby I'll pull in this PR so that you can rebase yours

PR #257, opened 2015-08-24
Add support for root distance and dispersion in NtpdCollector

jaingaurav:
These values aide in debugging ntp issues.


src/collectors/ntpd/ntpd.py
```diff
@@ -117,6 +117,30 @@ def get_ntpdc_stats(self):
                 data['max_error'] = val
             elif key == 'estimated error':
                 data['est_error'] = val
+            elif key == 'status':
+                data['status'] = val
+
+        return data.items()
+
+    def get_ntpdc_sysinfo_output(self):
+        return self.run_command([self.config['ntpdc_bin'], '-c', 'sysinfo'])
```
mattrobenolt: I'm not super familiar, but is there any case where this command may not exist or could fail?

jaingaurav: It's entirely possible, though I think unlikely if someone enabled this collector. It this an issue?

jaingaurav: Also note that the command already being used by this collector for the `kerninfo`

mattrobenolt: Sorry, I meant, is there a case where `ntpdc -c kerninfo` may work, but `ntpdc -c sysinfo` will not exist?

jaingaurav: From http://archive.ntp.org/ntp4/ChangeLog-stable if you search for `sysinfo` it seems that `sysinfo` and `kerninfo` were added at the same time. So the answer I guess is no, they should both exist.

mattrobenolt: :+1:

PR #258, opened 2015-08-25
Add systemd collector

gmlexx:
Send 1 if configured systemd service status is running and 0 otherwise

### DIFF
--- a/src/collectors/systemd/systemd-collector.py
+++ b/src/collectors/systemd/systemd-collector.py
@@ -31,6 +31,8 @@ class SystemdCollector(Collector):
 
     def collect(self):
 
+        manager = None
+
         try:
 
             manager = Manager()
@@ -50,4 +52,5 @@ class SystemdCollector(Collector):
 
         finally:
 
-            manager.unsubscribe()
+            if manager:
+                manager.unsubscribe()
--- a/src/collectors/systemd/systemd-collector.py
+++ b/src/collectors/systemd/systemd-collector.py
@@ -1,0 +1,53 @@
+# coding=utf-8
+
+"""
+Custom collector for systemd process control system
+(http://www.freedesktop.org/wiki/Software/systemd/)
+#### Dependencies
+ * python-systemd
+ * diamond
+#### Usage
+<pre>
+services = nginx,ntpd
+</pre>
+"""
+
+from diamond.collector import Collector
+from systemd.manager import Manager
+from systemd.exceptions import SystemdError
+
+
+class SystemdCollector(Collector):
+
+    def get_default_config(self):
+        """
+        Returns the default collector settings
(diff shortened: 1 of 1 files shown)


src/collectors/systemd/systemd-collector.py
```diff
@@ -0,0 +1,56 @@
...
+        Returns the default collector settings
+        """
+        config = super(SystemdCollector, self).get_default_config()
+        config.update({
+            'services': '',
+        })
+        return config
+
+    def collect(self):
+
```
jaingaurav: More of a style thing, please remove this empty line

src/collectors/systemd/systemd-collector.py
```diff
@@ -0,0 +1,56 @@
...
+        manager = None
+
+        try:
+
```
jaingaurav: More of a style thing, please remove this empty line

src/collectors/systemd/systemd-collector.py
```diff
@@ -0,0 +1,56 @@
...
+            manager = Manager()
+
+            for service in [s.strip() for s in self.config['services']]:
+
```
jaingaurav: More of a style thing, please remove this empty line

src/collectors/systemd/systemd-collector.py
```diff
@@ -0,0 +1,56 @@
...
+                try:
+                    unit = manager.get_unit('%s.service' % service)
+                    self.publish("%s.running" % service, 1
+                                 if str(unit.properties.SubState) == "running"
+                                 else 0)
+                except SystemdError, e:
+                    self.log.exception(e)
+
+        finally:
+
```
jaingaurav: More of a style thing, please remove this empty line

src/collectors/systemd/systemd-collector.py
```diff
@@ -0,0 +1,56 @@
+# coding=utf-8
+
+"""
+Custom collector for systemd process control system
+(http://www.freedesktop.org/wiki/Software/systemd/)
+#### Dependencies
+ * python-systemd
+ * diamond
```
jaingaurav: Diamond doesn't need to spelled out as a dependency here

jaingaurav:
@gmlexx So this looks reasonable however a few issues:

- No test cases
- I am confused about https://github.com/systemd/python-systemd vs https://github.com/wiliamsouza/python-systemd it seems the latter is no longer maintained. In fact the second last commit suggests using the official library. Is it possible to use the official library instead? Unfortunately when going through the docs, I could not see an obvious way to get the service stat information.

gmlexx:
Thank for review. The latter systemd module seems has some leak also.
I will look at the official module.

PR #259, opened 2015-08-26
Py3k add support for running the same codebase on python2.6/2.7 and 3.3+

Ormod:
This PR makes it possible to run and build Diamond with Python3. 

All unit tests on Python3 except for 3 (2 rabbitmq and 1 kafkatest) pass.
On Python2.x all the unittests pass as before.

I've tried to break up the changes into small logical, hopefully-easy-to-review incremental changes.

Note that this breaks compatility with Pythons before 2.6 but that has mostly been the de facto state for a lot of the collectors even before this. (they depend on things like subprocess.check_output)


Makefile
```diff
@@ -3,7 +3,8 @@ PROJECT=diamond
 VERSION :=$(shell bash version.sh )
 RELEASE :=$(shell ls -1 dist/*.noarch.rpm 2>/dev/null | wc -l )
 HASH	:=$(shell git rev-parse HEAD )
-DISTRO=precise
+DISTRO:=$(shell python -c "import platform;print(platform.linux_distribution()[0])")
+PYTHON?=python3
```
mattrobenolt: I assume you were using this for testing, but I'd rather us default any `PYTHON` to `python2`, or at least just `python` and let the environment gods use whichever is default.

bin/diamond-setup
```diff
@@ -77,7 +78,7 @@ def getCollectors(path):
 
 
 def typeToString(key):
-    if isinstance(obj.config[key], basestring):
```
mattrobenolt: Any of these `basestring` -> `str` translations have lost meaning.

This is covered in `six`, see: https://bitbucket.org/gutworth/six/src/8a545f4e906f6f479a6eb8837f31d03731597687/six.py?at=default#six.py-49

setup.py
```diff
@@ -17,12 +17,8 @@ def running_under_virtualenv():
     return False
 
 
-if os.environ.get('USE_SETUPTOOLS'):
```
mattrobenolt: Do you have any context on why this was a thing? Because I don't, which means, I'm a bit afraid to just drop it.

Is distutils not a thing at all in python3?

src/collectors/amavis/amavis.py
```diff
@@ -75,6 +75,8 @@ def collect(self):
                 cmdline = [self.config['amavisd_exe'], '-c', '1']
             agent = subprocess.Popen(cmdline, stdout=subprocess.PIPE)
             agent_out = agent.communicate()[0]
+            if isinstance(agent_out, bytes):
+                agent_out = agent_out.decode("utf8")
```
mattrobenolt: Why was this added?

mattrobenolt: Oh, TIL `bytes` is an alias for `str` in python2.

mattrobenolt: Is it safe to assume `utf8` conversion will always work? What about `ascii`?

src/collectors/cpuacct_cgroup/cpuacct_cgroup.py
```diff
@@ -62,8 +62,8 @@ def collect(self):
                 stat_file.close()
 
         # create metrics from collected utimes and stimes for cgroups
-        for parent, cpuacct in results.iteritems():
```
mattrobenolt: Can we keep the `iteritems` for python2? This is something handled by six.

src/collectors/darner/darner.py
```diff
@@ -137,7 +137,7 @@ def collect(self):
                             queues[queue][queue_stat])
 
             # figure out what we're configured to get, defaulting to everything
-            desired = self.config.get('publish', stats.keys())
+            desired = self.config.get('publish', list(stats.keys()))
```
mattrobenolt: We can normalize on using `iterkeys()` here.

src/collectors/disktemp/disktemp.py
```diff
@@ -67,7 +67,7 @@ def match_device(self, device, path):
             # If the regex has a capture group for pretty printing, pick
             # the last matched capture group
             if self.devices.groups > 0:
-                key = '.'.join(filter(None, [g for g in m.groups()]))
+                key = '.'.join([_f for _f in [g for g in m.groups()] if _f])
```
mattrobenolt: `if _f is not None` is more correct to the original meaning.

src/collectors/ip/ip.py
```diff
@@ -103,7 +104,7 @@ def collect(self):
             data = data.split()
 
             # Zip up the keys and values
-            for i in xrange(1, len(header)):
```
mattrobenolt: plz to use `six.moves.range` instead so we can preserve the same behavior on python2 and 3.

src/collectors/mongodb/test/testmongodb.py
```diff
@@ -6,13 +6,18 @@
...
-from mock import call
+from test import MagicMock
+from test import patch
+from test import call
 
 from diamond.collector import Collector
 from mongodb import MongoDBCollector
 
+try:
+    long
```
mattrobenolt: fwiw, in python2, there's literally no reason to ever use `long` afaik. `int()` will automatically coerce to a `long` if the value is long enough. So we can just stop using `long()` entirely.

src/collectors/netscalersnmp/netscalersnmp.py
```diff
@@ -215,8 +215,8 @@ def collect_snmp(self, device, host, port, community):
                                        community)[serviceTypeOid].strip("\'"))
 
             # Filter excluded service types
-            if serviceType in map(lambda v: int(v),
```
mattrobenolt: Why replace `map`?

src/collectors/ossec/ossec.py
```diff
@@ -53,7 +53,9 @@ def collect(self):
         try:
             p = subprocess.Popen(command, stdout=subprocess.PIPE)
             res = p.communicate()[0]
-        except Exception, e:
+            if isinstance(res, bytes):
+                res = res.decode("utf8")
+        except Exception as e:
             self.log.error('Unable to exec cmd: %s, because %s'
                            % (' '.join(command), str(e)))
```
mattrobenolt: Doesn't `str` here have different meaning between python2/3?

src/diamond/pycompat.py
```diff
@@ -0,0 +1,18 @@
+# coding=utf-8
```
mattrobenolt: We should just start by vendorizing the entirety of six for now imo. There are too many things we're using.

src/collectors/darner/darner.py
Ormod: Iterkeys does not exist in python3

mattrobenolt: Sorry, I mean, normalize on what six does. They provide a wrapper to do the same behavior in both.

src/collectors/ossec/ossec.py
Ormod: Not for the purposes of printing the exception name AFAIK.

josegonzalez:
Nice work!

varunpatro:
Any update on this? We use diamond heavily in our company and are now moving to python 3, so if I can help with this development, I'll be happy to help!

shortdudey123:
@Ormod can you rebase this PR and address some of the feedback?

Ormod:
I'm afraid we stopped using Diamond a while back but if someone else wants to work on this and re-use my code please feel free to do so.

Issue #183, opened 2015-05-09
RFC: Splitting out collectors


mattrobenolt:
I'm going to start by laying out the way collectors work now:

* collectors are bundled inside this https://github.com/python-diamond/Diamond repo, but are not installed onto a `PYTHONPATH` anywhere. Nor are they modules in the `diamond` namespace.
* collectors get installed by `setup.py` into their `collectors_path` which is something like `/usr/share/diamond/collectors/` on Debian/Ubuntu. See: https://github.com/python-diamond/Diamond/blob/master/setup.py and https://github.com/python-diamond/Diamond/blob/master/src/diamond/utils/classes.py#L100
* This is relatively similar to how Salt deals with extensions as well which also poses similar downsides

Pros:
* Diamond collectors are just a self contained directory that gets shoved onto the `collectors_path` without the need to be "installed".
* This nicely facilitates allowing other deb packages to install collectors into Diamond. (We do this at Disqus. See https://github.com/disqus/diamond-collectors)
* Makes system packaging pretty easy

Cons:
* They are all packaged into this one monolithic repo.
* We can't version individual collectors and maintain the outside of the core.
* Not very pythonic (they're not real modules or installable by `pip`)
* They require their own dependencies typically, which `Diamond` does not depend on, and there's no way to message this without reading the source of the module. See previous point.

Proposal (WIP):
* Handle the current process today (just crawl a directory of files)
* Handle _new_ style as plugins.
* Plugins get installed through pip. ala `pip install diamond-collector-something-awesome` which lists it's own dependencies
* Utilize `setuptools` and `entry_points` and declare a `diamond.collectors` namespace that plugins get registered into.
* Provide a template project that includes a setup.py/Makefile so the same repo can easily be packaged up properly for pip, and gets dropped into `collectors_path` for system packaging.

Thoughts?

/cc @josegonzalez

josegonzalez:
Sounds reasonable to me. As an alternative, here is [how python-beaver handled it](https://github.com/josegonzalez/python-beaver/pull/95/files#diff-6e04985bf14c2c1a4917cf3a38379a54R4). Your way looks neater though. Want to come up with a PR for this?

Investocat:
:+1:

redref:
:+1: 

May be a good way to re-launch "active" development on this project.

Seems quite in pause. Anything to help ?

mattrobenolt:
afaik all of the necessary internal work has been done and collectors can be split out whenever.

[...]

PR #263, opened 2015-08-31
Initial support for collectors being installed into an entry_point

mattrobenolt:
First thing needed for #183.

This introduces a `diamond.collectors` `entry_point`.

What this means is a package can have:

```python
setup(
  entry_points={
    'diamond.collectors': [
      'thing = thing',
    ],
  }
)
```

and when installed, Diamond will discover and import it like any of the other included collector modules.

As a proof of concept, I've created two projects that we'd leverage with this:

 * [pytest-diamond](https://github.com/python-diamond/pytest-diamond) pytest plugin that we'd leverage to run tests against standalone collector plugins
 * [diamond-redis](https://github.com/python-diamond/diamond-redis) The `redisstats` module extracted into a `diamond_redis` module, and installs into an `entry_point`.

### DIFF
--- a/src/diamond/utils/classes.py
+++ b/src/diamond/utils/classes.py
@@ -6,6 +6,7 @@ import sys
 import logging
 import inspect
 import traceback
+import pkg_resources
 
 from diamond.util import load_class_from_name
 from diamond.collector import Collector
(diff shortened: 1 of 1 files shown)


jaingaurav:
lgtm

PR #264, opened 2015-09-01
[memory] Allow to force psutil use.

Nihn:


src/collectors/memory/memory.py
```diff
@@ -59,7 +60,9 @@ def get_default_config(self):
         """
         config = super(MemoryCollector, self).get_default_config()
         config.update({
-            'path':     'memory',
+            'path': 'memory',
+            'method': 'Threaded',
+            'force_psutil': 'False'
```
josegonzalez: If this is true and `psutil` is false, we should probably at least log a warning or something.

src/collectors/memory/memory.py
```diff
@@ -81,8 +85,8 @@ def collect(self):
                     name = name.rstrip(':')
                     value = int(value)
 
-                    if ((name not in _KEY_MAPPING and
-                         'detailed' not in self.config)):
+                    if (name not in _KEY_MAPPING
+                            and 'detailed' not in self.config):
```
jaingaurav: So I just went through a whole PEP8 clean up and opted to use the double brackets instead of the offset statements. Is there a preference for this? I personally find this less readable.

jaingaurav:
So what's the general motivation of this diff? Was there an alternate code path if pustules was not present?

mwicat:
@jaingaurav The reason is that one may not want to depend on kernel readings and use psutil instead. Our case is that we have kernel version 3.2.0 (the newest one in debian wheezy) and it does not provide MemAvailable metric (https://linuxwithraj.wordpress.com/2015/03/10/memory-utilization-from-procmeminfo-memavailable/) .

jaingaurav:
@Nihn, @mwicat, I'd like to merge this PR in but its still pending the fixes that were asked for. Primarily:

1) Log a warning
2) Fix code indentation style

Also if you can put the reason for this change in the git commit detailed description, it would be much appreciated.

src/collectors/memory/memory.py
Nihn: There is error:
```
            if not psutil:
                self.log.error('Unable to import psutil')
                self.log.error('No memory metrics retrieved')
                return None
```

src/collectors/memory/memory.py
```diff
@@ -70,7 +73,8 @@ def collect(self):
         """
         Collect memory stats
         """
-        if os.access(self.PROC, os.R_OK):
+        if (os.access(self.PROC, os.R_OK) and
```
jaingaurav:
```
if ((os.access(self.PROC, os.R_OK) and
     self.config.get('force_psutil') != 'True')):
```

jaingaurav:
Please fix the last indentation and squash the commits so that it is a single commit and I can merge this in. Thanks for all your work.

jaingaurav:
Thanks, merged

Issue #266, opened 2015-09-02
collector_path in diamond.conf not working in current master


Slach:
after install latest master
diamond.log filled by 

<pre>
[2015-09-02 10:06:46,978] [MainThread] Can not find collector VMStatCollector 
[2015-09-02 10:06:46,978] [MainThread] Can not find collector RedisCollector 
[2015-09-02 10:06:46,979] [MainThread] Can not find collector NetworkCollector 
[2015-09-02 10:06:46,979] [MainThread] Can not find collector PostgresqlCollector 
[2015-09-02 10:06:46,979] [MainThread] Can not find collector CPUCollector 
[2015-09-02 10:06:46,979] [MainThread] Can not find collector MemoryCollector 
[2015-09-02 10:06:46,979] [MainThread] Can not find collector LoadAverageCollector 
[2015-09-02 10:06:46,979] [MainThread] Can not find collector NginxCollector 
[2015-09-02 10:06:46,981] [MainThread] Can not find collector DiskSpaceCollector 
[2015-09-02 10:06:46,981] [MainThread] Can not find collector DiskUsageCollector 
[2015-09-02 10:06:46,981] [MainThread] Can not find collector RabbitMQCollector 
</pre>

please see my comment
https://github.com/python-diamond/Diamond/commit/6f37763354df554dbfb475b1122a049a3534ddbe#diff-264916192cc0a7d2ccfa0c62630b34e7R102

i add some logging.info into utils/classes.py
and catch wrong login whitch always return True and skip add file to collector
<pre>
2015-09-02 10:06:43,968] [MainThread] os.path.join(path, f)=/usr/share/diamond/collectors/elasticsearch/elasticsearch.py != filter <built-in function filter>
</pre>

szop85:
Same here:

[2015-09-02 12:19:31,688] [MainThread] Can not find collector LoadAverageCollector
[2015-09-02 12:19:31,688] [MainThread] Can not find collector DiskSpaceCollector
[2015-09-02 12:19:31,688] [MainThread] Can not find collector DiskUsageCollector
[2015-09-02 12:19:32,691] [MainThread] Can not find collector VMStatCollector
[2015-09-02 12:19:32,691] [MainThread] Can not find collector CPUCollector
[2015-09-02 12:19:32,691] [MainThread] Can not find collector MemoryCollector
[2015-09-02 12:19:32,691] [MainThread] Can not find collector LoadAverageCollector
[...]

PR #267, opened 2015-09-02
Update classes.py

Slach:
remove built-in filter function usage
see https://github.com/python-diamond/Diamond/issues/266

### DIFF
--- a/src/diamond/utils/classes.py
+++ b/src/diamond/utils/classes.py
@@ -143,11 +143,11 @@ def load_collectors_from_paths(paths, file_filter=None):
                     collectors[key] = subcollectors[key]
 
             # Ignore anything that isn't a .py file
-            elif os.path.isfile(fpath) and \
-                            len(f) > 3 and \
-                            f[-3:] == '.py' and \
-                            f[0:4] != 'test' and \
-                            f[0] != '.':
+            elif (os.path.isfile(fpath) and
+                  len(f) > 3 and
+                  f[-3:] == '.py' and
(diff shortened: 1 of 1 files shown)


mattrobenolt:
Going to review today and make sure this is the correct fix.

Sorry about that. :(

jaingaurav:
Isn't the correct fix to re-add the lost `filter` parameter in `load_collectors()` and pass it to `load_collectors_from_paths()`

jaingaurav:
Also, clearly a test case was missing here :(

Slach:
i'm re-add filter parameter
but i don't have skills for add clear test case ;) sorry

jaingaurav:
So after taking a closer look. My suggestion for the latter changes was wrong. It doesn't seem like the `filter` parameter is being used anymore. I've manually merged the initial commit.

Thanks for the PR.

Tibingeo:
We are running a diamond collector to collect system metrics and show it on a graph. However, when I try to run the service using service diamond start, it will start but immediately get terminated. I tried to find the log. Syslog says init: diamond main process (32501) terminated with status 1

The following are the /etc/log/diamond/diamond.log

2016-09-11 08:07:25,649] [MainThread] Can not find collector AppMetricCollector

 [2016-09-11 08:07:29,713] [MainThread] Collector failed! Traceback (most recent call last): File "/usr/lib/pymodules/python2.7/diamond/utils/scheduler.py", line 73, in collector_process .....

PR #268, opened 2015-09-04
[libvirtkvm] convert option sort_by_uuid to a bool

kinhvan017:
Option sort_by_uuid need to be converted to a bool


mattrobenolt:
:+1: thx

PR #270, opened 2015-09-05
fix all occurences of using string as bool

hvnsweeting:


mattrobenolt:
This is fine. My only feedback is consistency between using `diamond.collector.str_to_bool`, and simply `str_to_bool`. I'd personally prefer `str_to_bool` only.

hvnsweeting:
@mattrobenolt I fixed that

jaingaurav:
Looks great. Thanks!

Issue #274, opened 2015-09-08
Diamond does not set groups properly when changing uid and gid


nhooey:
According to [Python issue 10032](http://bugs.python.org/issue10032), setting the `uid` and `gid` is not enough to simulate logging in as that user. The `os.initgroups()` function needs to be used to initialize the groups of the user being changed to, like this:

```python
    if gid != -1 and uid != -1:
        os.initgroups(pwd.getpwuid(uid).pw_name, gid)
```

If this is not done, Diamond can't invoke actions on files when it changes to a user that only has group permissions for those files, since Python doesn't know that it's a member of those groups.

nhooey:
This is fixed by #275.

jaingaurav:
PR has been merged, thanks

PR #275, opened 2015-09-08
Set groups when changing `uid` and `gid` since Python doesn't do it.

nhooey:
This fixes Diamond not being able to make use of group permissions on
files when the user is switched.

According to Python issue:
http://bugs.python.org/issue10032

The groups need to be manually set when changing the `uid`/`gid`
manually.


nhooey:
This fixes #274.

jaingaurav:
lgtm

PR #277, opened 2015-09-18
Use psutil.virtual_memory and swap_memory to collect memory usage

kinhvan017:
Use psutil.virtual_memory() and psutil.swap_memory() to monitor the actual memory usage

psutil.phymem_usage() and psutil.virtmem_usage() are deprecated.
Instead we now have psutil.virtual_memory() and psutil.swap_memory(),
which should provide all the necessary pieces to monitor the actual
system memory usage, both physical and swap/disk related.


jaingaurav:
lgtm, thanks!

PR #280, opened 2015-09-22
Ensure Solr URLs don't have multiple '/'s

zoidbergwill:
The URL formatting generally calls URLs with multiple forward slashes,
since the formatting includes a slash between host:port and path, while
the majority of the calls to `_get` also begin with forward slashes.

This can raise issues on some web servers.


jaingaurav:
Seems reasonable, merged

Issue #282, opened 2015-09-23
ElbCollector causes StatsdHandler to fail


mfox-socialware:
When using the ElbCollector in conjunction with the StatsdHandler a failure occurs that causes StatsdHandler to stop emitting metrics for all collectors. This message is logged:

> [2015-09-23 15:00:57,384] [MainThread] Traceback (most recent call last):
>  File "/usr/local/lib/python2.7/site-packages/diamond/handler/Handler.py", line 98, in _flush
>    self.flush()
>  File "/usr/local/lib/python2.7/site-packages/diamond/handler/stats_d.py", line 142, in flush
>    self._send()
>  File "/usr/local/lib/python2.7/site-packages/diamond/handler/stats_d.py", line 129, in _send
>    value = value - self.old_values[metric.path]
> TypeError: unsupported operand type(s) for -: 'NoneType' and 'NoneType'

Upon inspection of stats_d.py, it attempts to calculate a delta for non-gauge values so that it can send a counter (lines 125-136). The relative code is lines 127-130:

``` python
value = metric.raw_value
if metric.path in self.old_values:
    value = value - self.old_values[metric.path]
self.old_values[metric.path] = metric.raw_value
```

The TypeError we see in the log is related to the - operand on types None and None (line 129).

We know, based on this error, that metric.raw_value is NoneType, so I tracked down where that originates from within ElbCollector. Here is the method signature for publish_delayed_metric() (lines 183-206):

``` python
def publish_delayed_metric(self, name, value, timestamp, raw_value=None,
                           precision=0, metric_type='GAUGE',
                           instance=None):
```

None of the places where publish_delayed_metric is called in elb.py set raw_value to anything explicitly, so it is always going to be NoneType. This would not be a problem but some of the metrics sent by ElbCollector are of type COUNTER.

I would like to fix this and enter a pull request, but wanted some guidance as to the coding standards of the project. What is the difference between raw_value and value? If we have ElbCollector always set raw_value equal to value, would this fix the problem?

Thanks.

mfox-socialware:
Also, in testing stats_d.py, I believe the lines in question that calculate delta to be a logical error. Calculating a delta is actually behavior one would expect from a gauge, not from a counter. As we do not know when the flush interval of StatsD is, calculating the delta can cause negative values and other inaccuracies. I believe this code would work as desired if the delta calculations were removed entirely from the counter logic.

analogue:
[...]

PR #284, opened 2015-09-24
Fixed ElbCollector incompatibility with StatsdHandler

mfox-socialware:
The ElbCollector had an incompatibility with StatsdHandler that caused the handler to stop publishing metrics. This will fix Issue #282.


analogue:
:+1:

PR #285, opened 2015-09-24
Mesos CGroup collector

bmhatfield:
This collects cgroup statistics and translates them from task Run IDs into higher-order task name information. Because of the nature of the higher-order task naming, individual hostname information is not helpful, and is indeed harmful, so this moves the metrics gathered for this collector to the root of the graphite tree.


bmhatfield:
We've got this running in production, so I am going to pull this in. Happy to fix things up in the future.

PR #287, opened 2015-09-25
Add a collector to gather metrics from Loggly searches.

Jorge-Rodriguez:


Jorge-Rodriguez:
@jaingaurav Did you get the chance to peek at this other PR?

src/collectors/loggly/test/testloggly.py
```diff
@@ -0,0 +1,214 @@
...
+            'unauthorizedmetric.mock.field': 0,
+            'forbiddenmetric.mock.field': 0,
+            'gonemetric.mock.field': 0,
+            'serverrmetric.mock.field': 0,
+            'notimplementedmetric.mock.field': 0,
+            'throttledmetric.mock.field': 0,
+            'timeoutmetric.mock.field': 0,
+        }
+
+        for key, value in published_metrics.iteritems():
```
jaingaurav: You can simply use `assertPublishedMany`, have a look at `testsmart.py`

Jorge-Rodriguez: I know, but `assertPublishedMetricMany` resets the mock object after the assertion which breaks the following assertion of `assertUnpublishedMetricMany`.

test.py
```diff
@@ -195,24 +195,25 @@ def assertPublishedMetric(self, mock, key, value, expected_value=1):
...
-                                                  expected_value)
-
-        if precision is not None:
-            self.assertAlmostEqual(float(actual_value),
-                                   float(expected_value),
-                                   places=precision,
-                                   msg=message)
-        else:
-            self.assertEqual(actual_value, expected_value, message)
+        if expected_value:
```
jaingaurav: So this looks like a great catch, but I think the unpublished API is still a little bit busted in that it expects a dictionary. We should just make it take a list of keys. Nothing that necessarily needs to be done in this PR, as it would likely involve changing a bunch of test cases.

src/collectors/loggly/test/testloggly.py
jaingaurav: Interesting, so it seems that `testmountstats.py` simply calls `assertPublishedMany` followed by `assertUnpublishedMany`. Are you suggesting that that test is incorrect since it resets the mock in between?

If so, maybe its time we clean up these assertions and have a unified function called `assertMetrics(self, mock, published_dict, unpublished_list=None)`.

src/collectors/loggly/test/testloggly.py
```diff
@@ -0,0 +1,214 @@
...
+                'count': 'true',
+                'query': 'mock: gone'
+            },
+            'serverrmetric': {
+                'subdomain': 'mocksub',
+                'username': 'mockuser',
+                'password': 'mockpass',
+                'field': 'mock.field',
+                'count': 'true',
+                'query': 'mock: serverr'
```
jaingaurav: double r

src/collectors/loggly/test/testloggly.py
```diff
@@ -0,0 +1,214 @@
...
+class TestLogglyCollector(CollectorTestCase):
+    TEST_CONFIG = {
+        'interval': 10,
+        'metric': {
+            'testmetric': {
+                'subdomain': 'mocksub',
+                'username': 'mockuser',
+                'password': 'mockpass',
+                'field': 'mock.field',
+                'query': 'mock: query'
```
jaingaurav: Should these query params have whitespace?

src/collectors/loggly/loggly.py
```diff
@@ -0,0 +1,268 @@
...
+[ metric ]
+[[ test.metric ]]
+subdomain = example
+username = loggly_user
+password = loggly_pass
+field = some.field
+query = tag:some_tag AND some_text_field='VALUE'
+```
+
+Example config file LogglyCollector.conf using split configuration
```
jaingaurav: Is this a stale comment, since you took out the split configs?

src/collectors/loggly/loggly.py
```diff
@@ -0,0 +1,268 @@
...
+    def get_default_config_help(self):
+        """Update configuration help."""
+        config_help = super(
+            LogglyCollector, self).get_default_config_help()
+        config_help.update({
+            'id': 'Metric identifier',
+            'field': 'Name of the field to retrieve, required',
+            'count': 'Whether to retrieve the count of matching events',
+            'timestamp_field': 'Name of the timestamp field.'
+                               'Defaults to \'timestamp\'',
```
jaingaurav: Just for readability, does it make sense to double quote the string so you don't have to escape the single quote? If you choose to do so, please keep on the strings in the dictionary consistent, by making them use double quotes for the values (you don't have to change all the strings in the file).

src/collectors/loggly/loggly.py
```diff
@@ -0,0 +1,268 @@
...
+
+Example config file LogglyCollector.conf
+
+```
+enabled = true
+interval = 60
+
+[ metric ]
+[[ test.metric ]]
+subdomain = example
```
jaingaurav: Does it make sense to keep the subdomain, username and password as required for each metric? Shouldn't this be a global value for the collector instance? I'm not too familiar with loggly, so don't know if this makes sense. Remember that if multiple passwords are needed, a user could always instantiate another instance of the collector?

src/collectors/loggly/loggly.py
```diff
@@ -0,0 +1,268 @@
...
+
+        def retrieve_results(metric_config):
+            def now_as_timefield():
+                if metric_config.timestamp_format is 'int':
+                    return time.mktime(datetime.datetime.now().timetuple())
+                else:
+                    return datetime.datetime.now().strftime(
+                        metric_config.timestamp_format)
+
+            def log_error_and_continue(code):
```
jaingaurav: Can't you just use there response reason, or is there a difference I'm missing here.

http://www.python-requests.org/en/latest/api/#requests.Response.reason

src/collectors/loggly/loggly.py
```diff
@@ -0,0 +1,268 @@
...
+                                        'Not Implemented'),
+                    503: self.log.error('Loggly returned error 503:'
+                                        'Throttled'),
+                    504: self.log.error('Loggly returned error 504:'
+                                        'Gateway Timeout')
+                }
+                try:
+                    known_errors[code]
+                except KeyError:
+                    self.log.error('Loggly returned unrecognized error code %s'
```
jaingaurav: What value does having a set of recognized codes do?

jaingaurav:
Submitted a few comments, still need to look more. Looks like there may be some test framework rework desired. If so, perhaps lets keep those framework changes minimal in this PR, and you can perhaps submit a bug regarding the issues you mentioned where the mock is reset.

Also note, that before we merge this, it would be great to collapse a bunch of the diffs. Note that I would still keep the `test.py` you made as a separate commit.

src/collectors/loggly/test/testloggly.py
Jorge-Rodriguez: Intentional, though possibly misleading. It was intended as short for server error.

src/collectors/loggly/test/testloggly.py
Jorge-Rodriguez: Yes, loggly allows whitespace on query params, this is an indication of that capability, loosely replicating a loggly search query.

src/collectors/loggly/loggly.py
Jorge-Rodriguez: This is stale indeed, I'll take it out.

src/collectors/loggly/loggly.py
Jorge-Rodriguez: You're correct, and this set up is a side effect of the split configuration. I'll move this to the collector level.

src/collectors/loggly/loggly.py
Jorge-Rodriguez: Loggly has its own explanation for their error codes, most match the standard reason, but not necessarily, these come from their API documentation.

src/collectors/loggly/loggly.py
Jorge-Rodriguez: Same as above, this comes from Loggly's API documentation, there are the values they have special meanings for.

Jorge-Rodriguez:
I'm going to work on these comments and strip out the change to `test.py` onto a different PR. However the change to `test.py` is needed for the existing tests to pass.

src/collectors/loggly/loggly.py
jaingaurav: Alright, then how about we clean up this dictionary a little bit to only include the explanation. I feel like the `Loggly returned error 400:` prefix can be nicely abstracted out in line 185. Also I think you want a whitespace after the `:`

src/collectors/loggly/test/testloggly.py
jaingaurav: haha nice, lets make is `server error` since it doesn't seem to abbreviate any of the others.

jaingaurav:
Thanks for that. Yeah, just might be best to keep your `test.py` change as a separate commit and order it first, as your collector test suite depends on it.

src/collectors/loggly/test/testloggly.py
Jorge-Rodriguez: Weird, I just re-ran the test with `assertUnpublishedMetricMany` and it works as expected, but I most definitely saw the unexpected behaviour earlier. I presumed it was due to the mock object reset but it must have been some sort of side effect.

Jorge-Rodriguez:
@jaingaurav I don't know if you have any more insights on the latest changes, or should I squash the commits?

Jorge-Rodriguez:
Ooops, rebased to the wrong commit and mangled `test.py`, cherry picked python-diamond/Diamond#290 to fix.

src/collectors/loggly/loggly.py
```diff
@@ -0,0 +1,246 @@
...
+                'field': 'count',
+                'count': False,
+                'query': '*'
+            }
+            [config.update({key: value}) for key, value in defaults.items()
+                if key not in config.keys()]
+
+        super(LogglyCollector, self).process_config()
+
+        self.config_bunch = bunch.bunchify(self.config)
```
jaingaurav: So I'd prefer not using the bunch package to parse configs. None of the other collectors use it, thus leading to an inconsistency.

src/collectors/loggly/loggly.py
```diff
@@ -0,0 +1,246 @@
...
+            'subdomain': "Loggly subdomain to retrieve results from",
+            'query': "Loggly search query",
+            'username': "Loggly username",
+            'password': "Loggly password"
+        })
+        return config_help
+
+    def process_config(self):
+        """Prepare metrics' configuration."""
+        def set_defaults(config):
```
jaingaurav: Shouldn't you declare a `get_default_config()` function instead as we do with other collectors?

src/collectors/loggly/loggly.py
```diff
@@ -0,0 +1,246 @@
...
+                'field': bm[0],
+                'timestamp': parse_timestamp(metric_config, bm[1]),
+                'value': float(bm[2])}), bare_metrics)
+
+        def parallel_retrieval(results, metric, config):
+            results.update({metric: bunch.bunchify(
+                {'results': retrieve_results(config)})})
+
+        results = bunch.Bunch()
+        with concurrent.futures.ThreadPoolExecutor(
```
jaingaurav: This is a really nice feature to use from Python 3. I feel like other collectors should try and use it.

src/collectors/loggly/loggly.py
```diff
@@ -0,0 +1,246 @@
...
+                    503: 'Throttled',
+                    504: 'Gateway Timeout'
+                }
+                try:
+                    self.log.error('Loggly returned error %s: %s'
+                                   % (code, known_errors[code]))
+                except KeyError:
+                    self.log.error('Loggly returned unrecognized error code %s'
+                                   % code)
+                finally:
```
jaingaurav: Why do you expect the need for a `finally:` block? Can this be removed?

src/collectors/loggly/loggly.py
```diff
@@ -0,0 +1,246 @@
...
+                'events_URI': 'https://{0.subdomain}.loggly.com'
+                              '/apiv2/events?rsid=%s'.format(
+                                  self.config_bunch)})
+
+    def collect(self):
+        """Collect metrics from loggly."""
+        def parse_timestamp(metric_config, timestamp):
+            if metric_config.timestamp_format is 'int':
+                return datetime.datetime.fromtimestamp(
+                    timestamp / 10**(ceil(log10(timestamp)) - 10))
```
jaingaurav: Why is this needed? Can't you use `strptime` throughout?

src/collectors/loggly/loggly.py
jaingaurav: Also a comment linking to the documentation would be awesome

src/collectors/loggly/loggly.py
```diff
@@ -0,0 +1,246 @@
...
+            if search_response.status_code is 200:
+                events = requests.get(
+                    metric_config.events_URI % search_response
+                    .json()['rsid']['id'],
+                    auth=(self.config_bunch.username,
+                          self.config_bunch.password))
+
+                if events.status_code is 200:
+                    bare_metrics = [(metric_config.field,
+                                     event[metric_config.timestamp_field],
```
jaingaurav: So, looking at this, it seems you are trying to shove events into a stats framework. That doesn't seem correct. Usually you would put aggregations into a stats framework, such as the number of events in the last window.

src/collectors/loggly/loggly.py
Jorge-Rodriguez: These are metrics defaults not collector defaults, we need to process the actual configuration first.

src/collectors/loggly/loggly.py
Jorge-Rodriguez: `strptime` doesn't provide format codes for epoch offsets, which is why we need to differentiate between `int` and any other `strptime`-like formats. The math for the `int` type is there to ensure we're passing seconds from epoch as we can't assume the offset timestamps are specified in seconds or any other time subdivision.

src/collectors/loggly/loggly.py
Jorge-Rodriguez: This is exactly the purpose of this collector, to extract stats from numeric values on events.

Jorge-Rodriguez:
@jaingaurav Is there anything I missed on this PR? Any other improvements?

jaingaurav:
So I'm still a bit confused by the purpose of this collector. Typically collectors in diamond run periodically and then report a value based on the time they are run. Further handler back ends that I am used to typically expect values to be reported on a regular basis.

With this collector it seems you are periodically searching for events and then reporting them based on their time stamp. Neither the collector or the reporting to the handler seems to follow the model I've seen elsewhere. Does diamond really serve any purpose here other than being a cron-like daemon?

Jorge-Rodriguez:
This collector reports two types of metric, values extracted from numeric fields on loggly events (in which case it does work just like a cron-like daemon) and event count during the collector interval which is more like the usual model.

You could argue that yes, this collector leans on diamond to run a cron-like task, the advantage over cron however is centralization. What we are trying to achieve with this collector is extract metrics off of log events and place them on some sort of metric aggregator system, e.g. graphite. It stands to reason that one would not use Diamond just for this collector but in addition to an array of others which is why it makes sense, in my opinion, to be able to use Diamond for all of them, instead of maintaining a separate cron job to retrieve log event metrics.

In short, the way I see it, from an abstract point of view, all collectors grab numeric values from some source and publish them to a given target, which is what this collector does as well. The event search is limited to the interval of the collector so the reported metrics are artificially made diamond-like.

From another angle, the driving force for this collector was a request from project management. The managers said in these words: "We have the duration of postgres long running queries in our logs on loggly, we want to see that as a timeseries on grafana" and that's the reason this collector exists.

jaingaurav:
So I think regarding the event count, that is fine to keep in diamond. However I have found that trying to force event based data into a system like graphite only leads to trouble. Graphite doesn't really like holes in its time series.

My recommendation For events is to have a plugin in grafana. Grafana already has a plugin to add annotations with event-based data in Elastic Search. Perhaps you could extend that for your purposes.

jaingaurav:
@Jorge-Rodriguez Any update on this collector? Were you able to come up with an alternate design?

Jorge-Rodriguez:
@jaingaurav I did not manage to get an alternate design. Diamond is the only part of our pipeline that we actually own, everything else being hosted solutions so we don't really have many ways around it.

The collector has been working alright for us all this time, in any case.

jaingaurav:
@Jorge-Rodriguez thanks for the update. As mentioned, I'm happy to merge a collector which just takes the metrics. For events, if you still want to use diamond, I suggest you use the UserScript collector to publish them. That way you are free to publish events as you please and can also benefit the community with your event count collector.

Jorge-Rodriguez:
@jaingaurav Let me give it a think. I don't know if having the collector capture just the event count offers a lot of value over just doing everything with UserScript.
I'm going on holidays tonight, but I'll get back to you within a couple of weeks.

Jorge-Rodriguez:
@jaingaurav I took a look at this. For the time being we will continue to use this collector, but you're right that the UserScript collector could do the job, and with that in mind I don't think there's a lot of value for just the event count.

Feel free to close this PR if that's what you prefer.

jaingaurav:
Thanks for your work @Jorge-Rodriguez. Hopefully the comments helped. I'll close out this PR then.

PR #288, opened 2015-09-25
Tweaks to ProcessResources Collector

Jorge-Rodriguez:
This PR allows to split the configuration of this collector into separate files. This is convenient for environments managed by orchestration tools, where different services may want to plug themselves in to this collector.

This also allows to send the -1 value for the collector metrics when the collected process is not found, this makes it possible to detect dead services.

### DIFF
--- a/src/collectors/processresources/processresources.py
+++ b/src/collectors/processresources/processresources.py
@@ -197,9 +197,16 @@ class ProcessResourcesCollector(diamond.collector.Collector):
 
         # publish results
         for pg_name, counters in self.processes_info.iteritems():
-            metrics = (
-                ("%s.%s" % (pg_name, key), value)
-                for key, value in counters.iteritems())
+            if counters:
+                metrics = (
+                    ("%s.%s" % (pg_name, key), value)
+                    for key, value in counters.iteritems())
+            else:
+                if self.processes[pg_name]['count_workers']:
+                    metrics = (('%s.workers_count' % pg_name, 0), )
+                else:
+                    metrics = ()
+
             [self.publish(*metric) for metric in metrics]
             # reinitialize process info
             self.processes_info[pg_name] = {}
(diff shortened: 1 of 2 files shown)


jaingaurav:
So in diamond 4.0 you can run multiple instances of the ProcessResources collector. Why not do that instead of building up knowledge of split configs in the collector?

Also instead of the `report_missing` option, how about we add the pid as a metric and report the pid as -1 if a process is not found.

Jorge-Rodriguez:
@jaingaurav, that works as well, this PR was the in-house enhancement we wrote but I like your approach better, at least regarding the configuration.

I have doubts about tracking the pid for missing processes, because that may not be a metric one may be tracking in say, grafana. I'll be more inclined to leave the `report_missing` flag in and maybe just send the `worker_count` as 0 iif both options are set.

jaingaurav:
I think the `report_missing` flag is a bit verbose. Is there any reason not to always report the `workers_count` as 0 if nothing matches.

Jorge-Rodriguez:
I want to agree, I'm mostly concerned about keeping the original functionality of just not publishing anything if the process is missing.

jaingaurav:
It doesn't strike me to have any issues, as we would only be adding a metric as opposed to adding/renaming. Please let me know if you'd like to submit a PR. You can submit it under this PR, or start afresh with a new PR.

Jorge-Rodriguez:
Alright, I'll make these changes and push to the current branch so this PR gets updated.

jaingaurav:
Looks good. Could you please squash your commits?

Also, it would help to modify the test suite to include your change.

Jorge-Rodriguez:
@jaingaurav Test updated and commits squashed.

jaingaurav:
Thanks for the PR and for making the suggested changes

PR #290, opened 2015-09-28
Don't check values when checking unpublised metric

Jorge-Rodriguez:
Change as discussed on python-diamond/Diamond#287


jaingaurav:
Thanks for the great catch!

PR #293, opened 2015-10-01
ElasticSearch Handler support

imreFitos:
New ElasticSearchHandler that is compatible with the latest ES2.0 and uses the Elastic Beats format.


shawnbutts:
I don't see the need to rename the ES collector for this?

imreFitos:
The handler uses the elasticsearch package and the collector's name clobbers the path :( So it won't work unless the collector's filename is different. 

The collector class is still called the same so no configuration has to change.

src/diamond/handler/elasticsearchHandler.py
```diff
@@ -0,0 +1,232 @@
...
+        # attempt to connect
+        try:
+            url_list = []
+            for host in self.config['host_urls'].split(','):
+                url_list.append(host)
+            self.es = Elasticsearch(url_list)
+        except Exception as exc:
+            self.log.error("ElasticSearchHandler start failed: {}".format(exc))
+            self.enabled = False
+        # save options into vars
```
jaingaurav: Please spread out your code a little to make it easier to read. A newline above here might help.

src/diamond/handler/elasticsearchHandler.py
```diff
@@ -0,0 +1,232 @@
...
+        if not self.enabled:
+            return
+        if len(self.queue) == 0:
+            return
+        actions = []
+        action = None
+        current_timestamp = 0
+        current_collector = ''
+        current_host = ''
+        while len(self.queue) > 0:
```
jaingaurav: Newline above here

src/diamond/handler/elasticsearchHandler.py
```diff
@@ -0,0 +1,232 @@
...
+            self.flush_time = time.time()
+            self.flush()
+
+    def flush(self):
+        """Write all metrics to ES server"""
+        if not self.enabled:
+            return
+        if len(self.queue) == 0:
+            return
+        actions = []
```
jaingaurav: Newline

jaingaurav:
Don't have too much to add here aside from style suggestions. Could you please add some newlines/documentation to make the code more readable. Also, it would be best to also fix your commit history to have 2 distinct commits, one to move the collector and the other to add the handler.

imreACTmd:
All right, updated the commit history and fixed up the formatting according to PEP8, let me know what you think!

imreACTmd:
Any comments?

stuartbfox:
My comment is yes please! With grafana v2.5 now supporting native ES searching, it will greatly simplify things if we can output directly to ES

imreACTmd:
Hi, checking in again, is there something wrong with the commit?

jaingaurav:
Hi @imreACTmd looking at the commit history it doesn't seem like you have been able to squash it down to 2 commits as requested. Let me know if you need help with that.

jaingaurav:
Again, the reason for it to be as 2 commits as that it makes it a little bit easier to differentiate between the file move and the handler addition

imreFitos:
@jaingaurav yes please, I thought I managed to redo the commits but I obviously didn't succeed.

jaingaurav:
@imreACTmd @imreFitos http://stackoverflow.com/questions/9790448/how-to-update-a-pull-request should help you out.

lpic10:
Is someone working on this?

imreFitos:
I moved to Sumo Logic and gave up on this.

PR #294, opened 2015-10-06
Fix to preserve legacy mode on RabbitMQ collector across runs.

Jorge-Rodriguez:


jaingaurav:
Took a look at this PR and had trouble understanding the fix. Care to give further explanation?

Jorge-Rodriguez:
The metric paths on this collector are different depending on whether it's running on legacy mode or not.
Legacy mode is detected by the absence of a `vhosts` section in the configuration however, if in legacy mode, the collector introduces a `vhosts` section in the configuration programmatically and this causes the collector to be detected as non legacy on runs after the first, which in turn means that the metrics paths are different on every run after the first.

Removing the `vhosts` section after the run restores the legacy style configuration.

jaingaurav:
Thanks I understand the problem now. So though your fix is a fine solution, I think proper fix is to not modify the self.conf dictionary at all. Could you instead move the references to it to be a local variable like `vhost_conf`

Jorge-Rodriguez:
PR updated with the suggested fix.

jaingaurav:
Thanks for making the changes.

PR #303, opened 2015-10-13
Making diamond-setup sections clearer in output.

linsomniac:
This change makes the output in diamond-setup easier to see where a section is and what section you are on for each question.


jaingaurav:
lgtm

PR #305, opened 2015-10-15
Add support for MemoryCollector to collect percent

alde:
Also refactor the class slightly, breaking the branching off into two
separate methods.

### DIFF
--- a/src/collectors/memory/memory.py
+++ b/src/collectors/memory/memory.py
@@ -39,6 +39,9 @@ _KEY_MAPPING = [
     'VmallocUsed',
     'VmallocChunk',
     'Committed_AS',
+    'FreePercent',
+    'UsedPercent',
+    'MemUsed'
 ]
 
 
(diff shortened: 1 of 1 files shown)


jaingaurav:
@alde Which kernel/OS are you using? I don't see this on my /proc/meminfo Also can't seem ti see it in the linux sources:

http://lxr.free-electrons.com/source/fs/proc/meminfo.c

alde:
That's why it's being calculated by data that is available. I didn't want to do those calculations in Graphite, because I felt like it would make more sense to have them sent in.

jaingaurav:
Hrm, so I guess it's a bit odd because the _KEY_MAPPING list is used in the /proc/meminfo code path.

So my view on this is collectors should not try to compute values as this is something that graphite is quite capable of doing either with the carbon-aggregator or by having a query that performs the calculations. Diamond collectors should really try to send raw values wherever possible. This simplifies collectors and makes leads to less confusion regarding how a collector may have computed a value.

Please let me know if there is something I'm missing regarding your use case.

alde:
Well, we use Diamond to monitor mesos-slaves, so we try to aggregate memory usage using a query containing a wildcard for the server number, for example 'server.mesos-slave-*.memory.MemFree', to not have to make a hundred different entries manually.

Given that, graphite can't do calculations on matching names automatically (or at least not in any way that I've found).

jaingaurav:
@alde You might want to take look at the latest map/reduce functionality added in graphite:

http://graphite.readthedocs.org/en/latest/functions.html#graphite.render.functions.reduceSeries

Also, I believe if you put carbon-aggregator before your carbon-cache you may be able to perform this calculation.

alde:
Oh, thanks. I will check on that.

Thanks :)

jaingaurav:
@alde alright closing this PR for now. Please feel free to re-open or comment if you are still having issues.

Issue #306, opened 2015-10-15
when running tests with mock==1.3.0 some fail with AttributeError assert_called_once


GregMeno:
Looks like assert_called_once got removed in mock==1.1.0
It does work in mock==1.0.1
I recommend converting the tests to use assert_called_once_with

```
======================================================================
ERROR: test_get_port_stats (test_portstat.GetPortStatsTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/Diamond/env/local/lib/python2.7/site-packages/mock/mock.py", line 1298, in patched
    return func(*args, **keywargs)
  File "/root/Diamond/src/collectors/portstat/tests/test_portstat.py", line 67, in test_get_port_stats
    net_connections_mock.assert_called_once()
  File "/root/Diamond/env/local/lib/python2.7/site-packages/mock/mock.py", line 714, in __getattr__
    raise AttributeError(name)
AttributeError: assert_called_once

======================================================================
ERROR: test_process_file_should_work_with_real_service_perfdata (testnagiosperfdata.TestNagiosPerfdataCollector)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/Diamond/env/local/lib/python2.7/site-packages/mock/mock.py", line 1298, in patched
    return func(*args, **keywargs)
  File "/root/Diamond/src/collectors/nagiosperfdata/test/testnagiosperfdata.py", line 117, in test_process_file_should_work_with_real_service_perfdata
    publish_mock.assert_called_once()
  File "/root/Diamond/env/local/lib/python2.7/site-packages/mock/mock.py", line 714, in __getattr__
    raise AttributeError(name)
AttributeError: assert_called_once

----------------------------------------------------------------------
Ran 449 tests in 2.816s

FAILED (errors=2, skipped=4)
make: *** [test] Error 2
```

GregMeno:
I read the mock source and it seems there was never an assert_called_once.

using mock==1.0.1
I show with this patch that we were letting mock fool us:
https://github.com/python-diamond/Diamond/commit/35055076f67cf78be8a5769fa1c2f9a5cc9142fe

produces this failure:
```
======================================================================
ERROR: test_get_port_stats (test_portstat.GetPortStatsTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/root/Diamond/env/local/lib/python2.7/site-packages/mock.py", line 1201, in patched
    return func(*args, **keywargs)
[...]

PR #307, opened 2015-10-15
Wip work with mock 1.3.0

GregMeno:
You'll probably want me to squash this. I would be happy to.
Also I'm not sure about the usage of call_count. It seems that using assert_called_with is better.
What do you think?


GregMeno:
fixes #306

josegonzalez:
Needs a quick rebase and then we can merge :)

PR #308, opened 2015-10-15
make vmstat send the raw_value and precision properly

squirrely:
It was sending the raw_value as 2.


jaingaurav:
looks good, could you please squash the commits and then I'll pull it in?

Issue #219, opened 2015-07-08
SNMP collector not working 


harishva:
[root@graphite collectors]# cat SNMPInterfaceCollector.conf
enabled = True
path_suffix = ""
retries = 3
measure_collector_time = False
byte_unit = byte
timeout = 5

path = interface
interval = 60

[devices]
[fw01]]
host = 192.168.1.1
port = 161
community = public
# here  goes logs 

Traceback (most recent call last):
  File "/usr/lib/python2.6/site-packages/diamond/utils/scheduler.py", line 70, in collector_process
    collector._run()
  File "/usr/lib/python2.6/site-packages/diamond/collector.py", line 472, in _run
    self.collect()
  File "/usr/lib/python2.6/site-packages/diamond/collector.py", line 366, in collect
    raise NotImplementedError()
NotImplementedError
[2015-07-08 13:56:56,318] [MainThread] Collector failed!
Traceback (most recent call last):
  File "/usr/lib/python2.6/site-packages/diamond/utils/scheduler.py", line 70, in collector_process
    collector._run()
  File "/usr/lib/python2.6/site-packages/diamond/collector.py", line 472, in _run
    self.collect()
  File "/usr/lib/python2.6/site-packages/diamond/collector.py", line 366, in collect
    raise NotImplementedError()
NotImplementedError
[2015-07-08 13:56:58,332] [MainThread] Collector failed!
Traceback (most recent call last):
  File "/usr/lib/python2.6/site-packages/diamond/utils/scheduler.py", line 70, in collector_process
    collector._run()
  File "/usr/lib/python2.6/site-packages/diamond/collector.py", line 472, in _run
    self.collect()
  File "/usr/lib/python2.6/site-packages/diamond/collector.py", line 366, in collect
    raise NotImplementedError()
NotImplementedError

superjeb:
I'm having the same issue. 

Is there a fix?

superjeb:
NotImplementedError
[2015-07-30 22:17:29,256] [MainThread] Collector failed!
Traceback (most recent call last):
  File "/opt/diamond/lib/python2.7/site-packages/diamond/utils/scheduler.py", line 70, in collector_process
    collector._run()
  File "/opt/diamond/lib/python2.7/site-packages/diamond/collector.py", line 472, in _run
    self.collect()
  File "/opt/diamond/lib/python2.7/site-packages/diamond/collector.py", line 366, in collect
    raise NotImplementedError()
NotImplementedError
[...]

PR #63, opened 2015-01-14
Enhancements to SNMPCollector

shaunduncan:
In relation to Issue #59, I've taken the liberty to make some changes/enhancements based on some of the behavior I've seen. I couldn't tell if there was a major difference between the SNMPCollector and SNMPRawCollector, so I've combined that into just the base SNMPCollector. I've also included a large amount of tests with this as well. The changes here should be backwards compatible as some of the method signatures have not changed. I've also tested this on a VM running snmpd using both this improved SNMPCollector and the unmodified SNMPInterfaceCollector.

I've also added an enhancement to perform SNMP walks via a configuration. For example:

```
[oids]
1.2.3.* = foo.bar
```

Would produce metrics like `foo.bar.1` and `foo.bar.2.1.1.1`.

Please let me know if this change makes sense (or if it's just a terrible idea). I'd be happy to hear the feedback. Thanks!

PR #310, opened 2015-10-22
#219 - changing snmp base collector to actually implement collect

bmfurtado:
Unless I missed something terribly obvious the SNMP base collector (used by all the other SNMP collectors like SNMPRawCollector, SNMPInterfaceCollector, etc) never implemented the collect() method, resulting in a NotImplementedException every time you tried to use one of these.

It's not really obvious to me how this was ever working or how it got to this state... Perhaps there was an API change at some point in the past or something.

Nevertheless, I've quickly hacked a change that makes it work (though I haven't really tested it with all the SNMP collectors, nor have I written any tests for this) and I'd like to get some feedback :)

### DIFF
--- a/.travis.requirements.txt
+++ b/.travis.requirements.txt
@@ -19,3 +19,4 @@ simplejson
 setproctitle
 psycopg2
 PySensors
+pysnmp
--- a/src/collectors/snmp/snmp.py
+++ b/src/collectors/snmp/snmp.py
@@ -35,6 +35,10 @@ import diamond.collector
 
 class SNMPCollector(diamond.collector.Collector):
 
+    def __init__(self, *args, **kwargs):
+        super(SNMPCollector, self).__init__(*args, **kwargs)
+        self.snmpCmdGen = cmdgen.CommandGenerator()
+
     def get_default_config_help(self):
         config_help = super(SNMPCollector, self).get_default_config_help()
         config_help.update({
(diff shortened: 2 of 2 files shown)


bmfurtado:
Tests are failing due to lack of pysnmp... Will investigate later how other collectors are solving this problem with external dependencies.

jrpaz:
So we applied your fix and got the SNMPInterfaceCollector/SNMPRawCollector working but we have run into this issue.

https://github.com/BrightcoveOS/Diamond/issues/263

Any thoughts ?

bmfurtado:
@jrpaz reading through that issue it seems like it was never fixed (just acknowledged that it was a problem) so it doesn't surprise me that's still the case...

raphtheb:
I'm actually witnessing the same issue, it would be rather nice to have this fix merged in!

jaingaurav:
Seems like this might be a dup of PR #63, though that PR reaches further out

bmfurtado:
@jaingaurav I agree... I didn't notice that PR existed before I made this one... Feel free to close this one whenever you want assuming the other PR works (and eventually gets merged)

olivierHa:
Hello, 

@jaingaurav  Any updates on this issue ? snmpcollector aren't working for a while :(

jacksu:
why doesn‘t merge this request?

josegonzalez:
Merged, thanks for the pull request (and sorry for the wait).

PR #311, opened 2015-10-23
Ignore further SIGHUPs when handling it

cannium:
Otherwise a `SIGHUPException` would be raised and crash diamond


jaingaurav:
Seems reasonable, thanks

PR #312, opened 2015-10-25
Format metrics for InfluxDB 0.9

bluecamel:


jaingaurav:
@bluecamel This PR looks reasonable. Could you make the following modifications:

1) Squash all commits into a single commit (I'd like to have each commit pass unit tests independently)
2) Add test case for your change. Preferably one that only works after your change. Seems like the InfluxDB handle might be missing a test handler.
3) Add a description in the commit explaining the issue and how you fixed it.

Also does this PR result in InfluxDB 0.8 not longer being supported?

languitar:
Would be nice if this could get merged

shortdudey123:

PR #318, opened 2015-11-01
Pick up all metrics current Apaches expose

somechris:
The added metrics are exposed by current vanilla Ubuntu 14.04 LTS
Apaches, and can be seen live for example on

  http://www.apache.org/server-status?auto

(The EmptyWorkerSlot might seem pointless, but it's usefull in stacked graphs to show how much headroom one has when it comes to worker processes)


jaingaurav:
Seems reasonable. Could you please add a test case with a fixture demonstrating the new stats.

PR #320, opened 2015-11-05
Rate-limit reconnections

ggs-sglimm:
Hi,

we found that the GraphiteHandler (possibly others, too) can create quite a high network load, when the Carbon endpoint is down.

The current implementation tries to reconnect on every incoming metric. This causes both a TCP reconnection and (if no local DNS cache is used) a DNS query. This merge request rate limits the GraphiteHandler to once per 10 seconds (configurable).

### DIFF
--- a/src/diamond/handler/graphite.py
+++ b/src/diamond/handler/graphite.py
@@ -18,6 +18,7 @@ use it.
 
 from Handler import Handler
 import socket
+from time import time
 
 
 class GraphiteHandler(Handler):
@@ -49,7 +50,9 @@ class GraphiteHandler(Handler):
             self.config['trim_backlog_multiplier'])
         self.flow_info = self.config['flow_info']
         self.scope_id = self.config['scope_id']
+        self.reconnect_interval = self.config['reconnect_interval']
         self.metrics = []
+        self.lastConnectionAttempt = 0
 
         # Connect
         self._connect()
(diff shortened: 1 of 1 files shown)


jaingaurav:
Thanks for the PR. Few notes:
- Please squash these together into a single commit
- Please use snake_case instead of camelCase for function names and variables

Also, this seems to depend on the send() function to trigger the reconnect. Thus the reconnect may happen much later on if you are not constantly sending metrics. Did you see if there was better alternative to this?

ggs-sglimm:
>Please use snake_case instead of camelCase for function names and variables
Oops. Done.

>Please squash these together into a single commit
Done.

>Also, this seems to depend on the send() function to trigger the reconnect. Thus the reconnect may happen much later on if you are not constantly sending metrics. Did you see if there was better alternative to this?
As far as I can see, the only alternative would be a separate thread that handles reconnections - but I didn't want to introduce that complexity.

I would argue, that the old code had the same problem. Consider the following timeline:
- Carbon interface goes down
- Metrics are collected
- Carbon interface goes up again.
In this case, both the old and the new code will fail to send the metrics to carbon. The only difference exists, when carbon goes down / comes up _during_ a collection - which is unlikely, if your collections are infrequent.
In any case, you can get the old behaviour by setting reconnect_interval to 0.

shortdudey123:
Superseded by https://github.com/python-diamond/Diamond/pull/494

PR #321, opened 2015-11-05
Calamari

GregMeno:
This is a fork of the CephCollector that we've used in calamari since diamond3.4
I've rebased, cleaned up, fixed tests, and tested the package with real data
I look forward to discussing what needs to change in order to get this back upstream.

The main change here is to optionally roll up stats about a ceph cluster under one reporting structure instead of spreading it out under all the server nodes that comprise the cluster.

### DIFF
--- a/src/collectors/ceph/ceph.py
+++ b/src/collectors/ceph/ceph.py
@@ -83,7 +83,9 @@ def _popen_check_output(*popenargs):
     This is inspired by subprocess.check_output, added in Python 2.7. This
     method provides similar functionality but will work with Python 2.6.
     """
-    process = subprocess.Popen(*popenargs, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    process = subprocess.Popen(*popenargs,
+                               stdout=subprocess.PIPE,
+                               stderr=subprocess.PIPE)
     output, err = process.communicate()
     retcode = process.poll()
     if retcode:
@@ -99,7 +101,8 @@ class AdminSocketError(Exception):
         self.command = command
 
     def __str__(self):
-        return "Admin socket error calling %s on socket %s" % (self.command, self.socket_name)
+        message = "Admin socket error calling %s on socket %s"
+        return message % (self.command, self.socket_name)
 
 
 class MonError(Exception):
@@ -108,11 +111,13 @@ class MonError(Exception):
         self.command = command
 
     def __str__(self):
-        return "Mon command error calling %s on cluster %s" % (self.command, self.cluster_name)
+        message = "Mon command error calling %s on cluster %s"
+        return message % (self.command, self.cluster_name)
 
 
 class GlobalName(str):
     pass
+
 
 class LocalName(str):
     pass
@@ -121,11 +126,12 @@ class LocalName(str):
 class CephCollector(diamond.collector.Collector):
     def __init__(self, config=None, handlers=[], name=None, configfile=None):
         super(CephCollector, self).__init__(config, handlers, name, configfile)
-        self.config['short_names'] = str_to_bool(self.config['short_names'])
-        self.config['service_stats_global'] = str_to_bool(self.config['service_stats_global'])
-        self.config['perf_counters_enabled'] = str_to_bool(self.config['perf_counters_enabled'])
-        self.config['osd_stats_enabled'] = str_to_bool(self.config['osd_stats_enabled'])
-        self.config['long_running_detail'] = str_to_bool(self.config['long_running_detail'])
+        for key in ('short_names',
+                    'service_stats_global',
+                    'perf_counters_enabled',
+                    'osd_stats_enabled',
+                    'long_running_detail'):
+            self.config[key] = str_to_bool(self.config[key])
 
     def get_default_config_help(self):
         config_help = super(CephCollector, self).get_default_config_help()
@@ -140,14 +146,18 @@ class CephCollector(diamond.collector.Collector):
                            "in metric paths.  Defaults to true.",
             'cluster_prefix': "Prefix for per-cluster metrics.  Defaults"
                            "to 'ceph.cluster'.",
-            'osd_stats_enabled': "Whether to enable OSD service stats.  These are the most numerous and"
-                           "may overload underpowered graphite instances when there are 100s of OSDs. "
-                           "Defaults to true",
+            'osd_stats_enabled': "Whether to enable OSD service stats.  These"
+                           "are the most numerous and may overload"
+                           "underpowered graphite instances when there are "
+                           " 100s of OSDs. Defaults to true",
             'service_stats_global': "If true, stats from osds and mons are"
-                                    "stored under the cluster prefix (not by host).  If false, these"
+                                    "stored under the cluster prefix (not by"
+                                    "host).  If false, these"
                                     "stats are stored in per-host paths.",
-            'long_running_detail': "Whether to break down long running averages into sum/count/average (true), or"
-                                   "only output the average from the last measurement interval (false).  Defaults"
+            'long_running_detail': "Whether to break down long running"
+                                   "averages into sum/count/average (true), or"
+                                   "only output the average from the last"
+                                   "measurement interval (false).  Defaults"
                                    "to false."
         })
         return config_help
@@ -181,8 +191,9 @@ class CephCollector(diamond.collector.Collector):
         elif isinstance(name, LocalName):
             return super(CephCollector, self).get_metric_path(name, instance)
         else:
-            # Require explicit local or global indication to catch bugs more easily
-            raise RuntimeError("Name '{0}' not LocalName or GlobalName".format(name))
+            # explicit local or global indication to catch bugs more easily
+            message = "Name '{0}' not LocalName or GlobalName".format(name)
+            raise RuntimeError(message)
 
     def _get_socket_paths(self):
         """Return a sequence of paths to sockets for communicating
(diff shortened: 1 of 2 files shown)


src/collectors/ceph/ceph.py
```diff
@@ -12,22 +12,32 @@
...
 import subprocess
-
+import re
 import diamond.collector
+import diamond.convertor
+from diamond.collector import str_to_bool
+
+
+# Metric name/path separator
+_PATH_SEP = "."
```
jaingaurav: Good to make this a global, but there are a number of collectors that just simply use '.'

We should either convert everything to use a common variable, or use '.' throughout

src/collectors/ceph/ceph.py
```diff
@@ -69,78 +169,357 @@ def get_default_config(self):
...
-        publish the individual values.
+        :param val: string in format "seconds.nanoseconds" or
+                    floating point number.
+
+        Returns:
+            Time in seconds as a floating point number.
         """
+        if isinstance(val, basestring):
+            sec, nsec = map(lambda v: long(v), val.split("."))
+            return float(sec * _NSEC_PER_SEC + nsec) / float(_NSEC_PER_SEC)
```
jaingaurav: Can't you simply do `float(val)`

src/collectors/ceph/ceph.py
```diff
@@ -12,22 +12,32 @@
...
+# Metric name/path separator
+_PATH_SEP = "."
 
+_NSEC_PER_SEC = 1000000000
 
-def flatten_dictionary(input, sep='.', prefix=None):
-    """Produces iterator of pairs where the first value is
-    the joined key names and the second value is the value
-    associated with the lowest level key. For example::
+# Performance metric data types
```
jaingaurav: Could you link to the documentation with these values?

src/collectors/ceph/ceph.py
```diff
@@ -69,78 +169,357 @@ def get_default_config(self):
...
+        """Publish a set of Ceph performance counters, including schema.
+
+        :param prefix: string prefixed to metric names
+        :param stats: dictionary containing performance counters
+        :param schema: performance counter schema
+        """
+        for path, stat_type in flatten_dictionary(schema):
+            # remove 'stat_type' component to get metric name
+            if path[-1] != 'type':
+                continue
```
jaingaurav: Why do you skip over names which don't end in `type`

src/collectors/ceph/ceph.py
```diff
@@ -35,30 +45,120 @@ def flatten_dictionary(input, sep='.', prefix=None):
...
+        d = d[component]
+        element = d
+    return element
+
+
+class CalledProcessError(Exception):
+    pass
+
+
+def _popen_check_output(*popenargs):
```
jaingaurav: should we instead import subprocess32?

src/collectors/ceph/ceph.py
```diff
@@ -69,78 +169,357 @@ def get_default_config(self):
...
+                                                  '--admin-daemon',
+                                                  socket_path] + command)
+        except CalledProcessError, e:
+            raise AdminSocketError(socket_path, command)
+
+        try:
+            return json.loads(json_blob)
+        except (ValueError, IndexError):
+            self.log.exception('Error parsing output from %s' % socket_path)
+            raise AdminSocketError(socket_path, command)
```
jaingaurav: Should we make this a single try/catch with 2 except statements?

src/collectors/ceph/ceph.py
```diff
@@ -35,30 +45,120 @@ def flatten_dictionary(input, sep='.', prefix=None):
...
+    retcode = process.poll()
+    if retcode:
+        msg = "Command '%s' exited with non-zero status %d" % \
+              (popenargs[0], retcode)
+        raise CalledProcessError(msg)
+    return output, err
+
+
+class AdminSocketError(Exception):
+    def __init__(self, socket_name, command):
```
jaingaurav: Please add brief documentation for the error

src/collectors/ceph/ceph.py
```diff
@@ -35,30 +45,120 @@ def flatten_dictionary(input, sep='.', prefix=None):
...
+        self.socket_name = socket_name
+        self.command = command
+
+    def __str__(self):
+        message = "Admin socket error calling %s on socket %s"
+        return message % (self.command, self.socket_name)
+
+
+class MonError(Exception):
```
jaingaurav: Please add brief documentation for the error

src/collectors/ceph/ceph.py
```diff
@@ -69,78 +169,357 @@ def get_default_config(self):
...
+            'long_running_detail': False,
+            'perf_counters_enabled': True
         })
         return config
 
+    def get_metric_path(self, name, instance=None):
+        """
+        This collector returns some cluster-wide statistics rather than
+        server-specific statistics, so we override this to
+        avoid diamond prefixing the hostname to our metrics.
```
jaingaurav: Where does diamond prefix the hostname?

src/collectors/ceph/ceph.py
GregMeno: @jaingaurav I am happy to change this and I'm not quite sure I understand your concern here. _PATH_SEP is "private" to this module and currently set to the same token other collectors use. It's only really a problem if someone goes and changes it. Are you suggesting that I pull this into a separate PR and we can discuss it later? I don't really have strong feelings one way or the other.

src/collectors/ceph/ceph.py
GregMeno: @jaingaurav I would prefer not doing that. Popen can raise ValueError when called with invalid args. It would make the self.log.exception misleading

src/collectors/ceph/ceph.py
GregMeno: I will.

src/collectors/ceph/ceph.py
GregMeno: I will. Happy to

src/collectors/ceph/ceph.py
GregMeno: not sure. I'll have to dig in to figure it out. 
@jcsp Do you have any insight here?
On first reading it must be that ceph is expressing time in a seconds.nanoseconds and if we just used float(val) we'd get an incorrect coercion based on seconds.deci_centi_milli_.._nano. Does that make sense?

jcsp: The comment says it's seconds.nanoseconds, so I'm inclined to believe it (don't remember writing this myself)

josegonzalez:
I thought we were going to try and separate out the collectors into different packages?

mattrobenolt:
> I thought we were going to try and separate out the collectors into different packages?

This technically works. :)

See: https://github.com/python-diamond/diamond-redis/blob/master/setup.py#L23-L27

When I last did this, it fully worked. This was just never documented or "released" as a feature.

src/collectors/ceph/ceph.py
GregMeno: I think it's reasonable. pypi lists subprocess32 as available for python 2.4-2.7. It would mean adding it to setup.py
https://github.com/python-diamond/Diamond/blob/master/setup.py#L146

Does that sound like the right thing to do?

GregMeno:
@josegonzalez @mattrobenolt I would be happy to separate the CephCollector into it's own package. Is this something I should consider doing in this pull request?

src/collectors/ceph/ceph.py
GregMeno: Done

src/collectors/ceph/ceph.py
GregMeno: The perf counter schema http://docs.ceph.com/docs/master/dev/perf_counters/#schema actually contains elements beyond type e.g. 
{u'description': u'a version',
 u'nick': u'',
 u'type': 2}
seems like I need to get the ceph docs updated to reflect reality

We skip the other elements for two reasons 
1. it would result in us publishing individual stats more than once for each collection 
see: https://github.com/GregMeno/Diamond-1/blob/calamari/src/collectors/ceph/test/testceph.py#L225

2. we need the type value from the schema to determine how to publish the stat

src/collectors/ceph/ceph.py
GregMeno: https://github.com/python-diamond/Diamond/blob/master/src/diamond/collector.py#L343

src/collectors/ceph/ceph.py
GregMeno: done

jaingaurav:
There is still quite a bit of work to do before we can break up the collectors into separate packages, so I consider this pull request still valid. However this PR demonstrates the motivation for the separation, i.e. for many collectors, the diamond maintainers are not in a great position to comment.

I'll try to take a look more next week as I plough through the outstanding PRs.

GregMeno:
@jaingaurav Thank you for setting my expectations. I appreciate your effort to review this.

GregMeno:
@jaingaurav @josegonzalez @mattrobenolt I'd like to pick this discussion back up. What can I do to get some or all of this upstream? Should I be working to package the collector? Is that something gets into distros that ship diamond? Thank you for your time.

src/collectors/ceph/ceph.py
jaingaurav: So my main concern is that this is private at all. Technically the way a metric is separated might mot be the role of the collector, but in fact the handler. It's just odd that a specific collector is choosing to define this as opposed to it being a global in the base class.

src/collectors/ceph/ceph.py
jaingaurav: @GregMeno could you please make this change then?

jcsp: I don't think there's a change needed here.  The input value is in <seconds>.<nanoseconds> format.  So if you had "1.1", that would *not* be equal to 1.1 seconds.  It would be equal to 1.000000001 -- the existing code is correct.

josegonzalez:
Sorry, I've been busy with quite a few other OSS projects. I'll take a look at this in a bit.

shortdudey123:
@GregMeno can you squash your commits a bit?
@josegonzalez any thoughts? this looks fine to me

mvdkleijn:
Hey guys, hope this doesn't come over as pushy (not the intention!) :smile: 

Any chance this PR can get some lovin'? It seems to be blocking several things downstream which in turn is causing users (of Calamari) headaches who're trying to update it to upgrade Ceph to Infernalis or Jewel+

shortdudey123:
I will take a look at carrying these commits into a new PR that has them squashed down

shortdudey123:
@ChristinaMeno Can you rebase this on master?  Then I will merge it.

PR #327, opened 2015-11-19
statsd: use pystatsd pipeline to send batches

filippog:
the pipeline feature has been introduced in https://github.com/jsocol/pystatsd
since version 2.0

### DIFF
--- a/src/diamond/handler/stats_d.py
+++ b/src/diamond/handler/stats_d.py
@@ -135,6 +135,8 @@ class StatsdHandler(Handler):
                     statsd.Counter(prefix, self.connection).increment(
                         name, value)
 
+        if hasattr(statsd, 'StatsClient'):
+            self.connection.send()
         self.metrics = []
 
     def flush(self):
@@ -152,7 +154,7 @@ class StatsdHandler(Handler):
             self.connection = statsd.StatsClient(
                 host=self.host,
                 port=self.port
-            )
+            ).pipeline()
         else:
             # Create socket
             self.connection = statsd.Connection(


src/diamond/handler/stats_d.py
```diff
@@ -135,6 +135,8 @@ def _send(self):
                     statsd.Counter(prefix, self.connection).increment(
                         name, value)
 
+        if hasattr(statsd, 'StatsClient'):
```
jaingaurav: What is the purpose of this check?

src/diamond/handler/stats_d.py
```diff
@@ -152,7 +154,7 @@ def _connect(self):
             self.connection = statsd.StatsClient(
                 host=self.host,
                 port=self.port
-            )
+            ).pipeline()
```
jaingaurav: Is there documentation for this feature? Does this call need to be conditional for legacy support?

filippog: yup documented at http://statsd.readthedocs.org/en/v3.2.1/pipeline.html for https://pypi.python.org/pypi/statsd/ and introduced in version 2.0 according to jsocol/pystatsd@51d684e46d

afaict this collector supports both the above and https://pypi.python.org/pypi/python-statsd/1.7.2 with the former being actively maintained on github but not the latter, hence the conditional

src/diamond/handler/stats_d.py
filippog: conditional to support both python-statsd and statsd, see also other comment

src/diamond/handler/stats_d.py
jaingaurav: Is there a reason to support both versions of the library? Seems like the former offers a superior API.

filippog: besides backward compatibility? no idea if there is still a good reason nowadays, possibly not

jaingaurav:
Thanks for the clarification. Perhaps in a future diamond version we can deprecate support for the older library

filippog:
yup, I can followup with PRs to change python-statsd with statsd in tox/travis/etc

PR #331, opened 2015-11-27
suggest statsd in place of python-statsd

filippog:
first step to stop support for python-statsd
(https://github.com/WoLpH/python-statsd) in place of
(https://github.com/jsocol/pystatsd)


jaingaurav:
Looks good, so is the idea that we move over all test cases, but leave the handler as is?

filippog:
yeah I think it makes sense to start defaulting to statsd first, I've just updated the PR to include also a deprecation warning in the handler itself too

jaingaurav:
Looks good.

Issue #271, opened 2015-09-07
Postgres password required in diamond config file


dvarrazzo:
The postgres collector has a `postgres` default string for the password connection. This basically forces to override the password by writing it in clear text in the `diamond.conf` file or to use a weak auth mode such as  `trust` in `pg_hba.conf`.

There should be a way to not pass a password to the psycopg2 `connect()` function so that the password can be specified in the `.pgpass` file.

jaingaurav:
@dvarrazzo seems like a reasonable suggestion. Any chance you'd be able to provide a PR? Otherwise you might get luck pining some of the dev who have worked on that collector.

bmhatfield:
Fixed by #332

PR #332, opened 2015-11-29
Feature/postgres collector auth with pgpass

kAlmAcetA:
Option to authenticate to postgres with password stored in ~/.pgpass.
Added new config option `password_provider` that can take values:
- `password` - default, auth with supplied password from config
- `pgpass` - use ~/.pgpass

Using pgpass is implemented in libpq, pyscopg2 is based on, it only requires to not provide password.


kAlmAcetA:
It is related to  #271

bmhatfield:
Awesome, thanks!

PR #334, opened 2015-12-02
Added ability to send Cloudwatch Metrics to AWS/AutoScaling namespace

six0h:
This patch affects the Cloudwatch Handler. It adds the ability to send metrics to the AWS/AutoScaling namespace by defining the AWS/AutoScaling Namespace, and grabbing the "aws:autoscaling:groupName" from the Instance Tags.

The original Cloudwatch handler automatically assigned all metrics to the instance that they were sent from. This adds the ability to send any given metric to an AutoScaling group as well, as a metric for the user to scale on.

### DIFF
--- a/src/diamond/handler/cloudwatch.py
+++ b/src/diamond/handler/cloudwatch.py
@@ -156,7 +156,7 @@ class cloudwatchHandler(Handler):
             self.cloudwatch = boto.ec2.cloudwatch.connect_to_region(
                 self.region)
             self.log.debug(
-                "CloudWatch: Succesfully Connected to CloudWatch at Region: %s",
+                "CloudWatch: Succesfully Connected to CloudWatch Region: %s",
                 self.region)
         except boto.exception.EC2ResponseError:
             self.log.error('CloudWatch: CloudWatch Exception Handler: ')
@@ -210,7 +210,9 @@ class cloudwatchHandler(Handler):
                 )
                 try:
                     if self.autoscaling:
-                        self.log.debug('Cloudwatch: Publishing metric to AutoScaling')
+                        self.log.debug(
+                            'Cloudwatch: Publishing metric to AutoScaling'
+                        )
                         self.cloudwatch.put_metric_data(
                             str(rule['namespace']),
                             str(rule['name']),
--- a/src/diamond/handler/cloudwatch.py
+++ b/src/diamond/handler/cloudwatch.py
@@ -189,7 +189,9 @@ class cloudwatchHandler(Handler):
 
         try:
             if self.config['autoscaling']:
-                instances = self.ec2.get_only_instances(["%s" % self.instance_id])
+                instances = self.ec2.get_only_instances([
+                    "%s" % self.instance_id
+                ])
                 inst = instances[0]
                 self.log.debug(inst.tags['aws:autoscaling:groupName'])
                 autoscaling_group = inst.tags['aws:autoscaling:groupName']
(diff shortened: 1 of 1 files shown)


src/diamond/handler/cloudwatch.py
```diff
@@ -170,6 +181,11 @@ def process(self, metric):
         metricname = str(metric.getMetricPath())
         timestamp = datetime.datetime.fromtimestamp(metric.timestamp)
 
+        if self.autoscaling:
+            instance = self.ec2.get_only_instances([self.instance_id])
+            self.log.debug(instance[0].tags['aws:autoscaling:groupName'])
+            autoscaling_group = instance[0].tags['aws:autoscaling:groupName']
```
bmhatfield: Should this have a guard against the tag possibly missing, to avoid breaking the handler invocation with `KeyError`s?

src/diamond/handler/cloudwatch.py
```diff
@@ -80,7 +83,10 @@ def __init__(self, config=None):
         self.log.debug("Setting InstanceId: " + self.instance_id)
 
         self.valid_config = ('region', 'collector', 'metric', 'namespace',
-                             'name', 'unit')
+                             'name', 'unit', 'autoscaling')
+
+        if self.config['autoscaling']:
+            self.autoscaling = True
```
bmhatfield: Why assign this to `self.autoscaling` instead of just pulling it from the config when needed?

src/diamond/handler/cloudwatch.py
```diff
@@ -193,12 +209,24 @@ def process(self, metric):
...
+                    if self.autoscaling:
+                        self.log.debug(
+                            'Cloudwatch: Publishing metric to AutoScaling'
+                        )
+                        self.cloudwatch.put_metric_data(
+                            str(rule['namespace']),
+                            str(rule['name']),
+                            str(metric.value),
+                            timestamp, str(rule['unit']),
+                            {'AutoScalingGroupName': autoscaling_group})
```
bmhatfield: It seems to me that these statements are essentially identical, with the exception of this line. Perhaps it makes sense to DRY this up a little by doing something like:

```python
if self.autoscaling:
  target = {'AutoScalingGroupName': autoscaling_group}
else:
  target = {'InstanceId': self.instance_id}

self.cloudwatch.put_metric_data(
                            str(rule['namespace']),
                            str(rule['name']),
                            str(metric.value),
                            timestamp, str(rule['unit']),
                            target)
```

And you can add your debug logging about where we're publishing to the simpler `if`.

src/diamond/handler/cloudwatch.py
six0h: Thanks for your comments! Good thought, I think the error should occur, but it should be a more obvious error that tells the user why the key doesn't exist. If they're trying to report to autoscaling on an instance that is not in autoscaling, then we should tell them thats what they've done.

src/diamond/handler/cloudwatch.py
six0h: fair enough, will fix.

src/diamond/handler/cloudwatch.py
six0h: great idea. will fix.

six0h:
Added ability to provide autoscaling group name and instance ID to cloudwatch from config file. Calling AWS from multiple instances to get the instance id and autoscaling information caused us to hit API Throttling limits pretty quickly.

six0h:
Is there any way I can assist further in getting these changes integrated?

masonleung:
Any update on merging this?

josegonzalez:
This needs tests in order to be merged.

six0h:
There is a way to easily achieve the same effect, built in to cloudwatch today. I am closing this pull request, it is now redundant.

PR #336, opened 2015-12-02
Ignore warnings for lsof -bu

MichaelDoyle:
Suppress warnings when running `lsof -bu <user>`. 

Per the man pages

```
-b	causes lsof to avoid  kernel  functions	 that  might  block  -
		lstat(2), readlink(2), and stat(2).
```

Unfortunately, this causes warnings to be printed stdout, which is undesirable.

```
lsof: avoiding readlink(/): -b was specified.
lsof: avoiding stat(/): -b was specified.
lsof: WARNING: can't stat() rootfs file system /
      Output information may be incomplete.
```

### DIFF
--- a/src/collectors/filestat/filestat.py
+++ b/src/collectors/filestat/filestat.py
@@ -236,7 +236,7 @@ class FilestatCollector(diamond.collector.Collector):
         d = {}
         for u in users:
             d[u] = {}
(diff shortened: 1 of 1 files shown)


jaingaurav:
lgtm thanks

PR #338, opened 2015-12-03
Collect cpuacct.usage for mesos cgroup

ajorgensen:
There is a bug in ubuntu that causes significant drift between
cpuacct.usage and cpuacct.stats
(https://bugs.launchpad.net/ubuntu/+source/linux-lts-trusty/+bug/1497447).
This changes adds the collection of cpuacct.usage because it more
accurately reflects how much cpu time has actually been used.


bmhatfield:
Looks great, thanks!

PR #339, opened 2015-12-04
Adding number of CPU cores

thardie:
Added a metric under cpu total "ncores" which records the number of
cores on this machine. This is useful if trying to divide any total
values by the number of CPU cores to get an actual percentage

### DIFF
--- a/src/collectors/memcached/memcached.py
+++ b/src/collectors/memcached/memcached.py
@@ -87,13 +87,26 @@ class MemcachedCollector(diamond.collector.Collector):
             else:
                 sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
                 sock.connect((host, int(port)))
+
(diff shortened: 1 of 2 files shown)


jaingaurav:
Thanks for the PR. I would request the same improve as I did for PR# 162. Please take a look and see what you can do.

Also I'd appreciate squashing the commits to logical chunks once your done.

src/collectors/nfacct/nfacct.py
```diff
@@ -22,7 +22,7 @@ def get_default_config_help(self):
         config_help = (
             super(NetfilterAccountingCollector, self).get_default_config_help())
         config_help.update({
-            'bin': 'The path to the smartctl binary',
+            'bin': 'The path to the nfacct binary',
```
shortdudey123: Unrelated?

PR #340, opened 2015-12-09
Missing %s for string interpolation

ajorgensen:
Without the `s` it was expecting a number and caused it to crash with:

```
  File "/usr/lib/pymodules/python2.7/diamond/utils/scheduler.py", line 73, in collector_process
    collector._run()
  File "/usr/lib/pymodules/python2.7/diamond/collector.py", line 472, in _run
    self.collect()
  File "/usr/share/diamond/collectors/mesos_cgroup/mesos_cgroup.py", line 85, in collect
    with open(os.path.join(task_id, "%.usage" % aspect)) as f:
TypeError: %u format: a number is required, not str
~
```


bmhatfield:
Thanks. I really need to get these tests uploaded, but I am not going to block merging this on that.

PR #341, opened 2015-12-09
Remove undefined key variable from cpuacct uage

ajorgensen:


bmhatfield:
And, of course, tests would have caught this. We'll pull them into this PR before merging.

bmhatfield:
Fixed in #342

PR #350, opened 2015-12-16
Add openvz support

badele:


badele:
I have tested in production environment with Proxmox server, it  work fine :)

shortdudey123:
I can do the doc file separately

PR #353, opened 2015-12-19
Removed '.' prefix from metric_name

sbasgall:
Removed '.' prefix from metric_name which causes the metric path to have an empty node.

### DIFF
--- a/src/collectors/nfsd/nfsd.py
+++ b/src/collectors/nfsd/nfsd.py
@@ -190,7 +190,7 @@ class NfsdCollector(diamond.collector.Collector):
             file.close()
 
             for stat in results.keys():
-                metric_name = '.' + stat
+                metric_name = stat
                 metric_value = long(float(results[stat]))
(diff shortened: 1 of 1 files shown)


eserte:
Currently the situation is that nfsd metrics are rejected if blacklisting with the default blacklist.conf is enabled in graphite-carbon. See: https://github.com/graphite-project/carbon/blob/master/conf/blacklist.conf.example#L7

eserte:
The same problem existed in the nfs (client) collector, but was fixed in https://github.com/python-diamond/Diamond/commit/d888277601efc965fbff4d6a9b80c4f55ce1ce37

shortdudey123:
@sbasgall can you fix the failing test?

eserte:
@shortdudey123 I think the test fix is in https://github.com/python-diamond/Diamond/issues/354

shortdudey123:
hmm odd place to put it, but ok

shortdudey123:
Superceeded by #362

PR #354, opened 2015-12-19
Removed '.' prefix from keys in metrics dictionary

sbasgall:
Removed '.' prefix from keys in metrics dictionary to update the tests for the nfsd collector with respect to #353


jaingaurav:
Shouldn't PR #353 and this one be part of a single PR? Also I'm very surprised this was setup like this. Any idea why the prefix was put?

sbasgall:
I agree they should be combined into a single PR. I didn't notice the tests also needed to be updated until a later point in time and I wanted to get the conversation started in case I was completely missing something.

I can only speculate as to why it is like this.

My guess is that at one point the publish() call didn't provide a trailing '.' when joining the metric name to the metric path prefix and hostname and this collector wasn't updated when that change was made.

I haven't done any investigation to see if that is actually the case.

A commit that encompasses this same change to nfs.py ( which is written in the same general style ) was done last year - SHA: d888277601efc965fbff4d6a9b80c4f55ce1ce37

Should I go ahead and submit a combined PR?

jaingaurav:
Him, would it instead make sense to take the full pull request:

https://github.com/BrightcoveOS/Diamond/pull/637

If seems a serious issue is that the incorrect column.

shortdudey123:
Superceeded by #362

PR #358, opened 2015-12-22
Support amzn image

jumping:
On the amazon images could not install by `python setup.py install`, the reason is the platform package could not detect the distribution from the platform.dist() which return 

```
>>> platform.dist()
('', '', '')
```


bmhatfield:
Looks like setup.py needs some PEP8 cleanup.

Also, can you add the logrotate config in a separate PR from this one? These contexts seem to be unrelated.

setup.py
```diff
@@ -59,6 +63,8 @@ def running_under_virtualenv():
                            glob('conf/collectors/*')))
         data_files.append(('/etc/diamond/handlers',
                            glob('conf/handlers/*')))
+        data_files.append(('/etc/logrotate.d',
+	                   ['debian/logrotate/diamond']))
```
jaingaurav: I think you need to add another space here to fix the PEP8 issues.

jaingaurav:
@bmhatfield Is the logrotate rule needed? I thought we manage log rotation ourselves ing the diamond config?

bmhatfield:
@jaingaurav it may not be, which is one reason why I suggested separating the concern into a separate PR

jumping:
@bmhatfield @jaingaurav Thank you.

jumping:
I created another pull request https://github.com/python-diamond/Diamond/pull/360 for support amz, but for the logrotate function decide to not pull now.

shortdudey123:
Closing since https://github.com/python-diamond/Diamond/pull/360 was merged and log rotate.d config is not needed

create a new PR if this is needed in the future

PR #362, opened 2015-12-24
Removed '.' prefix from metric name and formatted proc columns

sbasgall:
This pull request applies changes from d888277 to the nfsd collector.
It is a combination of PRs #353 and #354.


jaingaurav:
@sbasgall Thanks for the PR, however I think i'm still unclear on the way forward here.

Would it instead make sense to take the full pull request:

BrightcoveOS/Diamond#637

If so, then I can pull those relevant commits from the BrightCoveOS repo

sbasgall:
The PR BrightcoveOSDiamond#367 is for a different collector. If you were to accept that PR it would change the files in the nfs collector not the nfsd collector. 

The files are almost similar enough that you could apply the same patch against the file this PR changes but the proc file is parsed looking for different values in this collector.

sbasgall:
Is there any update on this PR? I would be happy to answer any additional questions.

As the nfsd collector exists right now, it is in a broken state.

This is apply a similar change that was done to a different colletor ( nfs ) awhile back but the fix was never applied to this collector.

PR #363, opened 2015-12-25
Fio collector

jordant:
Runs fio , the flexable i/o tester, and publishes results. Little different from other collectors as we actually generate some load and collect metrics on the results. If this is way outside the scope of what this project wants to do, I totally understand.

### DIFF
--- a/src/collectors/fio/fio.py
+++ b/src/collectors/fio/fio.py
@@ -14,23 +14,32 @@ try:
 except ImportError:
     import simplejson as json
 
-import glob
-import os
 import re
 import subprocess
 
 import diamond.collector
 
 
-def flatten_dict(d):
-    def items():
-        for key, value in d.items():
-            if isinstance(value, dict):
-                for subkey, subvalue in flatten_dict(value).items():
-                        yield key + "." + subkey, subvalue
-            else:
+def flatten_dictionary(input, sep='.', prefix=None):
+    """Produces iterator of pairs where the first value is
+    the joined key names and the second value is the value
+    associated with the lowest level key. For example::
+
+      {'a': {'b': 10},
+       'c': 20,
+       }
+
+    produces::
+
+      [('a.b', 10), ('c', 20)]
+    """
+
+    for key, value in sorted(input.items()):
+        if isinstance(value, dict):
+            for subkey, subvalue in flatten_dictionary(value):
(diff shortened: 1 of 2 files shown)


jaingaurav:
Interesting choice of a collector. Could you please describe an example of where this may be useful? I just want to make sure the underlying behavior of diamond actually suits your objective.

src/collectors/fio/fio.py
```diff
@@ -0,0 +1,121 @@
...
+            'directory': '/tmp',
+        })
+        return config
+
+    def _get_stats_from_fio(self):
+        """Return the parsed JSON fio output
+        """
+
+        try:
+            json_blob = subprocess.check_output(
```
mattrobenolt: Worth noting that check_output only exists in python 2.7, so this won't work in 2.6.

src/collectors/fio/fio.py
```diff
@@ -0,0 +1,121 @@
...
+                ['/usr/bin/fio',
```
mattrobenolt: Typically you wouldn't want to hardcode this location either, since someone may have it installed into `/usr/local/bin` instead, or any number of other places. So I'd move this to a configurable option, and have this as a default.

jordant:
Sorry it took me so long to get back on this. So basically this is a collector that uses fio ( http://linux.die.net/man/1/fio ) to test disk performance and report the results back to diamond. This makes this module unusual when compared to other diamond modules, in that it will actually generate load , by running fio. 

My main motivation for creating this was fio-graphite ( https://github.com/jordant/fio-graphite ) which takes fio output and ships to graphite. We do quite a bit of performance testing at DreamHost and diamond collectors make it very easy to get results from those load tests and store in graphite.

src/collectors/fio/fio.py
```diff
@@ -0,0 +1,121 @@
...
+except ImportError:
+    import simplejson as json
+
+import re
+import subprocess
+
+import diamond.collector
+
+
+def flatten_dictionary(input, sep='.', prefix=None):
```
shortdudey123, line 23: Looks like you copied this code and changed it.  Any reason you no longer use prefix in the code? See ceph collector code.

src/collectors/fio/fio.py
shortdudey123: This still needs to be dealt with.

PR #364, opened 2016-01-04
Fix RabbitMQ collector with recent change in rabbitmq-management

jippi:
Due to https://github.com/rabbitmq/rabbitmq-management/pull/74 and https://github.com/rabbitmq/rabbitmq-management/issues/100

`consumer_utilisation` used to be exposed as `""` when empty, its now `null` / `None` - causing it to no be emitted with a default value of `0` no more

I applied this fix for our own code, but I'm not a python developer, so there may be a more smooth way to do it

### DIFF
--- a/src/collectors/rabbitmq/rabbitmq.py
+++ b/src/collectors/rabbitmq/rabbitmq.py
@@ -238,13 +238,21 @@ class RabbitMQCollector(diamond.collector.Collector):
         if isinstance(value, dict):
             for new_key in value:
                 self._publish_metrics(name, keys, new_key, value)
-        elif isinstance(value, (float, int, long)):
+        elif isinstance(value, (float, int, long, type(None))):
             joined_keys = '.'.join(keys)
+
             if name:
                 publish_key = '{0}.{1}'.format(name, joined_keys)
             else:
                 publish_key = joined_keys
+
             if isinstance(value, bool):
                 value = int(value)
+
+            if isinstance(value, type(None)):
+                if "consumer_utilisation" in publish_key:
+                    value = 0
+                else:
+                    return False
 
             self.publish(publish_key, value)


ankr:
:+1:

jippi:
bump?

src/collectors/rabbitmq/rabbitmq.py
```diff
@@ -238,13 +238,21 @@ def _publish_metrics(self, name, prev_keys, key, data):
...
                 publish_key = joined_keys
+
             if isinstance(value, bool):
                 value = int(value)
 
+            if isinstance(value, type(None)):
+                if "consumer_utilisation" in publish_key:
+                    value = 0
+                else:
+                    return False
```
jaingaurav: Why are you returning `False` here? it doesn't seem like this function usually returns any value.

src/collectors/rabbitmq/rabbitmq.py
```diff
@@ -238,13 +238,21 @@ def _publish_metrics(self, name, prev_keys, key, data):
...
             if name:
                 publish_key = '{0}.{1}'.format(name, joined_keys)
             else:
                 publish_key = joined_keys
+
             if isinstance(value, bool):
                 value = int(value)
 
+            if isinstance(value, type(None)):
+                if "consumer_utilisation" in publish_key:
```
jaingaurav: Why special case for `consumer_utlization` we should make it universal right?

src/collectors/rabbitmq/rabbitmq.py
```diff
@@ -238,13 +238,21 @@ def _publish_metrics(self, name, prev_keys, key, data):
...
-        elif isinstance(value, (float, int, long)):
+        elif isinstance(value, (float, int, long, type(None))):
             joined_keys = '.'.join(keys)
+
             if name:
                 publish_key = '{0}.{1}'.format(name, joined_keys)
             else:
                 publish_key = joined_keys
+
             if isinstance(value, bool):
```
jaingaurav: How would this if condition ever occur?

src/collectors/rabbitmq/rabbitmq.py
```diff
@@ -238,13 +238,21 @@ def _publish_metrics(self, name, prev_keys, key, data):
         if isinstance(value, dict):
             for new_key in value:
                 self._publish_metrics(name, keys, new_key, value)
-        elif isinstance(value, (float, int, long)):
+        elif isinstance(value, (float, int, long, type(None))):
```
jaingaurav: What if we just did `value = data.get(key, 0)` on line 236? Is there a problem where we'd be reporting too many empty keys?

jaingaurav:
@jippi: Some comments for your consideration.

josegonzalez:
@jippi *bump*

jippi:
i don't use diamond any more, so the PR should be considered stale :) its almost 1y old anyway

josegonzalez:
@jippi I'll buy you a beer at the next cakefest? :P

If not, give me commit bit on your fork and I'll fix it.

PR #370, opened 2016-01-09
-add nvidia gpu collector

WeiBanjo:


src/collectors/nvidia_gpu/nvidia_gpu.py
```diff
@@ -0,0 +1,69 @@
+# coding=utf-8
+
+"""
+The NvidiaGPUCollector collects GPU utilization metric using nvidia-smi.
+
+#### Dependencies
+
+ * /usr/bin/nvidia-smi
```
jaingaurav: Could you please provide a link in the comments

src/collectors/nvidia_gpu/nvidia_gpu.py
```diff
@@ -0,0 +1,69 @@
...
+
+import diamond.collector
+
+
+class NvidiaGPUCollector(diamond.collector.ProcessCollector):
+
+    def get_default_config_help(self):
+        config_help = super(NvidiaGPUCollector, self).get_default_config_help()
+        config_help.update({
+            'bin': 'The path to the ping binary'
```
jaingaurav: Comment probably meant nvidia-smi binary

src/collectors/nvidia_gpu/nvidia_gpu.py
```diff
@@ -0,0 +1,69 @@
...
+        Collector GPU stats
+
+        """
+
+        stats_config = self.config['stats']
+        raw_output = self.run_command([
+            '--query-gpu={query_gpu}'.format(query_gpu=','.join(stats_config)),
+            '--format=csv,nounits,noheader'
+        ])
+        if raw_output is None:
```
jaingaurav: Could you please add a blank line above here for readability?

src/collectors/nvidia_gpu/nvidia_gpu.py
```diff
@@ -0,0 +1,69 @@
...
+            return
+        results = raw_output[0].strip().split("\n")
```
jaingaurav: Could you please add a blank line above here for readability?

src/collectors/nvidia_gpu/nvidia_gpu.py
```diff
@@ -0,0 +1,69 @@
...
+        for result in results:
+            stats = result.strip().split(',')
+            assert len(stats) == len(stats_config)
+            index = stats[0]
+            for stat_name, metric in izip(stats_config[1:], stats[1:]):
+                metric_name = 'gpu_{index}.{stat_name}'.format(
```
jaingaurav: Should the metric name be prefixed with gpu_ ? Is this a common naming convention or can there be other types of objects?

src/collectors/nvidia_gpu/nvidia_gpu.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def get_default_config(self):
+        """
+        Returns the default collector settings
+        """
+        config = super(NvidiaGPUCollector, self).get_default_config()
+        config.update({
+            'path': 'nvidia',
+            'bin': '/usr/bin/nvidia-smi',
+            'stats': [
```
jaingaurav: Could you add a help description for this?

jaingaurav:
Thanks for the collector. Please try to address the comments and we can merge it in.

src/collectors/nvidia_gpu/nvidia_gpu.py
WeiBanjo: Header from `nvidia-smi` has different formats. It is hard to convert into metric name.

src/collectors/nvidia_gpu/nvidia_gpu.py
WeiBanjo: The prefix is determine which GPUs the stats are coming from. For example, an instance has 2 GPUs, the stats will prefix with `gpu_0` and `gpu_1`.

I am following the pattern how CPUCollector publishes stats for each CPU.

src/collectors/nvidia_gpu/nvidia_gpu.py
jaingaurav: Does it make more sense then to use the python bindings instead?

http://pypi.python.org/pypi/nvidia-ml-py/

src/collectors/nvidia_gpu/nvidia_gpu.py
jaingaurav: So looking at the python library there seems to be a `nvmlDeviceGetName` function. Wouldn't that be a more consistent naming choice?

WeiBanjo: `nvmlDeviceGetName` returns the model name of the GPU instead of the index.

src/collectors/nvidia_gpu/nvidia_gpu.py
WeiBanjo: I would prefer to use `nvidia-smi` command line tool. Because Nvidia driver provides `nvidia-smi` tool and I don't have to install extra dependency.

Also the adoption of the library is really low, I am hesitate to use that library.

WeiBanjo:
Are we good on this PR ?

shortdudey123:
:+1: on getting this merged :)

jaingaurav:
@WeiBanjo So though I'm by no means knowledgable on this library. From what I'm reading, the python binding are supported and used by components such as the Ganglia plugin: https://developer.nvidia.com/ganglia-monitoring-system.

Thus it seems to me that we should try to use those binding rather than calling the tool directly. It just seems easier for people to install the pypi package rather than the command line tool.

WeiBanjo:
I am still trying to figure out how to mock those nvidia library methods like `nvmlDeviceGetCount ` and `nvmlDeviceGetHandleByIndex`. 

if you can give me a hint, I can add couple more tests.

PR #371, opened 2016-01-11
Google pubsub handler

dcsquared13:
Handler added to publish to google pub/sub.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -3,6 +3,7 @@ Copyright (C) 2011-2012 by Ivan Pouzyrevsky
 Copyright (C) 2011-2012 by Rob Smith            http://www.kormoc.com
 Copyright (C) 2012 Wijnand Modderman-Lenstra    https://maze.io/
 Copyright (C) 2012 Dennis Kaarsemaker           <dennis@kaarsemaker.net>
+Copyright (C) 2016 Google Inc.                  https://www.google.com/
 
 License: MIT
 
--- a/src/diamond/handler/pubsub.py
+++ b/src/diamond/handler/pubsub.py
@@ -156,7 +156,6 @@ class PubsubHandler(Handler):
 
                 self.metrics.append(tmp_msg)
 
-
     def _convert_to_pubsub(self, metric):
         """
         Convert a metric to a dictionary representing a Pub/Sub event.
--- a/src/diamond/handler/pubsub.py
+++ b/src/diamond/handler/pubsub.py
@@ -47,7 +47,8 @@ It has these options:
                   to send in 64K increments.
  * `batch_size` - If msgs are to be batched this will contain either the
                   count number or size in bytes.
-
+ * `tags` - Comma separated free-form field for additional key/value pairs
+            to be sent.
 """
 
 from Handler import Handler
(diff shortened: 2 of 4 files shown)


src/diamond/handler/pubsub.py
```diff
@@ -0,0 +1,216 @@
+# coding=utf-8
+
+"""
+Send metrics to [Google Pub/Sub](https://cloud.google.com/pubsub/).
+
+#### Copyright
+
+Copyright 2016 Google Inc.
```
josegonzalez: Why is the handler copyrighted to google?

dcsquared13: Because I work for Google so anything I contribute back has to have this included.

dcsquared13: I talked to legal here to make sure I'm speaking correctly.  I can remove the copyright verbiage from the pubsub.py file but it needs to stay in the main LICENSE.  A direct quote from legal is:  "This is not because Google endorses the project per se, but because we're a copyright author.  Brightcove Inc. is already listed in the LICENSE, the the project has demonstrated an understanding of corporate copyright authorship you can point to."

Hope that clears it up.

LICENSE
```diff
@@ -3,6 +3,7 @@ Copyright (C) 2011-2012 by Ivan Pouzyrevsky
 Copyright (C) 2011-2012 by Rob Smith            http://www.kormoc.com
 Copyright (C) 2012 Wijnand Modderman-Lenstra    https://maze.io/
 Copyright (C) 2012 Dennis Kaarsemaker           <dennis@kaarsemaker.net>
+Copyright (C) 2016 Google Inc.                  https://www.google.com/
```
dcsquared13: See response to Jose listed above.

src/diamond/handler/pubsub.py
shortdudey123: FYI... Brightcove Inc is the origional author, not a contributor

shortdudey123: See https://github.com/python-diamond/Diamond#repos

kormoc: FWIW, I'm a-okay with this. Having a copyright listed only really will pertain if we ever re-license away from the MIT license we're currently under. If we re-license, we would have to seek approval from every contributor anyway.

josegonzalez:
@dcsquared13 not that we have great handler test coverage, but I would super appreciate it if you had time to do that. If not, I'm fine with a merge - so long as another maintainer gives this a 👍.

shortdudey123:
@josegonzalez you want to chime in w/ your thoughts on the license as well? (adding google and mixing licenses)

josegonzalez:
IANAL but seems reasonable to let them sublicense their specific bit as whatever. I'm assuming a google lawyer was like "seems legit".

shortdudey123:
I agree w/ the handler side, but not quite sure about the copyright on the whole project from a legal stand point

hy777369-maker (commented):
Hello friends

hy777369-maker:
You helping

PR #377, opened 2016-01-20
Jolokia collector fixes

kasprzol:
We have encountered some problems with the Jolokia collector frequently timing out. The root cause was Jolokia's degraded performance when a connection wasn't closed (probably caused by signal in Diamond's scheduler interrupting the `urllib2.urlopen` call).

This pull request fixes this root cause (by adding timeouts to HTTP requests) as well as some of the slowdowns (e.g. caches regular expressions and the list of JMX domains).

### DIFF
--- a/src/collectors/jolokia/jolokia.py
+++ b/src/collectors/jolokia/jolokia.py
@@ -60,7 +60,6 @@ import urllib2
 class JolokiaCollector(diamond.collector.Collector):
 
     LIST_URL = "/list"
-    READ_URL = "/?ignoreErrors=true&p=read/%s:*"
 
     """
     These domains contain MBeans that are for management purposes,
@@ -88,7 +87,16 @@ class JolokiaCollector(diamond.collector.Collector):
             'port': 'Port',
             'rewrite': "This sub-section of the config contains pairs of"
                        " from-to regex rewrites.",
-            'path': 'Path to jolokia.  typically "jmx" or "jolokia"'
+            'path': 'Path to jolokia.  typically "jmx" or "jolokia"',
+            # https://github.com/rhuss/jolokia/blob/959424888a82abc2b1906c60547cd4df280f3b71/client/java/src/main/java/org/jolokia/client/request/J4pQueryParameter.java#L68
+            'use_cannonical_names': 'Whether property keys of ObjectNames'
+                                    ' should be ordered in the canonical way'
+                                    ' or in the way that they are created. The'
+                                    ' allowed values are either "True" in'
+                                    ' which case the canonical key order (=='
+                                    ' alphabetical sorted) is used or "False"'
+                                    ' for getting the keys as registered.'
+                                    ' Default is "True',
         })
         return config_help
 
@@ -103,6 +111,7 @@ class JolokiaCollector(diamond.collector.Collector):
             'password': None,
             'host': 'localhost',
             'port': 8778,
+            'use_cannonical_names': True,
         })
         return config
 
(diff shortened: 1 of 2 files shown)


bmhatfield:
cc @MichaelDoyle to take a look here.

@kasprzol - can you check out why CI is failing?

kasprzol:
@bmhatfield: too much rebasing, too litle sleep :) should be fixed now.

src/collectors/jolokia/jolokia.py
```diff
@@ -119,53 +163,75 @@ def check_mbean(self, mbean):
...
+                for k, v in mbeans.iteritems():
+                    if self.check_mbean(k):
+                        self.collect_bean(k, v)
 
     def read_json(self, request):
         json_str = request.read()
         return json.loads(json_str)
 
-    def list_request(self):
+    def _list_request(self):
```
MichaelDoyle: If we're going to change this here, can we change the other "private" methods to use the "_" convention as well? We should just be consistent one way or another.

src/collectors/jolokia/jolokia.py
```diff
@@ -119,53 +163,75 @@ def check_mbean(self, mbean):
...
+        """Retruns a dictionary with JMX domain names as keys"""
```
MichaelDoyle: Spelling - should be "Returns" instead of "Retruns". :)

MichaelDoyle:
I think these all look like good changes. Thanks! Just made a couple small comments.

src/collectors/jolokia/jolokia.py
```diff
@@ -83,7 +87,16 @@ def get_default_config_help(self):
             'port': 'Port',
             'rewrite': "This sub-section of the config contains pairs of"
                        " from-to regex rewrites.",
-            'path': 'Path to jolokia.  typically "jmx" or "jolokia"'
+            'path': 'Path to jolokia.  typically "jmx" or "jolokia"',
+            # https://github.com/rhuss/jolokia/blob/959424888a82abc2b1906c60547cd4df280f3b71/client/java/src/main/java/org/jolokia/client/request/J4pQueryParameter.java#L68
+            'use_cannonical_names': 'Whether property keys of ObjectNames'
```
MichaelDoyle: Spelling `use_cannonical_names` should be `use_canonical_names`

src/collectors/jolokia/jolokia.py
```diff
@@ -98,20 +111,51 @@ def get_default_config(self):
             'password': None,
             'host': 'localhost',
             'port': 8778,
+            'use_cannonical_names': True,
```
MichaelDoyle: Spelling `use_cannonical_names` should be `use_canonical_names`

src/collectors/jolokia/jolokia.py
```diff
@@ -72,6 +72,10 @@ def get_default_config_help(self):
         config_help = super(JolokiaCollector,
                             self).get_default_config_help()
         config_help.update({
+            'domains_to_check': "Pipe delimited list of JMX domains from which"
```
MichaelDoyle: We can probably simply name this `domains`.

src/collectors/jolokia/jolokia.py
```diff
@@ -119,53 +163,75 @@ def check_mbean(self, mbean):
...
-            return self.read_json(response)
-        except (urllib2.HTTPError, ValueError):
-            self.log.error('Unable to read JSON response.')
+            # https://jolokia.org/reference/html/protocol.html
+            # The optional parameter maxDepth can be used to restrict the depth
+            # of the return tree. Two value are possible: A maxDepth of 1
+            # restricts the return value to a map with the JMX domains as keys,
+            # a maxDepth of 2 truncates the map returned to the domain names
+            # (first level) and the MBean's properties (second level). The final
+            # values of the maps don't have any meaning and are dummy values.
```
MichaelDoyle: We can probably simplify this comment to explain why we've chosen maxDepth=1 and refer to the doc for the rest of the details/options.

MichaelDoyle:
I think these all look like good changes. Thanks! Just made a couple small comments.

kasprzol:
Thanks for the feedback. I've fixed the issues you highlighted. I also added an option to separate the path component of metrics from the path component of the jolokia url.

kasprzol:
Hi

Are you planning on merging this in? Or do you have any other comments?

bmhatfield:
@MichaelDoyle let me know if you think this is all good to go, and I can merge it.

MichaelDoyle:
LGTM :key: :sparkles:

PR #379, opened 2016-01-21
-add kafka consumer lag collector

WeiBanjo:
This collector is going to collect Kafka consumer lag metrics using `bin/kafka-run-class.sh kafka.tools.ConsumerOffsetChecker`


josegonzalez:
Why isn't this just part of the existing KafkaCollector?

WeiBanjo:
Because KafkaCollector is using JMX metrics and JMX doesn't expose Kafka consumer metrics.

PR #381, opened 2016-01-26
Add docker stats collector

csabiftw:
Added a new collector for docker using the stats api.


bmhatfield:
This looks simple, great and useful. Although there's some overlap with cgroup-style collectors, I think this is enough of a value-add by looking through the docker API instead (and has more docker-specific information) that it's worth including.

Thanks for this, and thanks for including simple and good tests!

bmhatfield:
Dang, just noticed there's already a docker collector that does almost the same thing. Can you help me understand the value of this second one? Perhaps the first can be enhanced?

I am going to revert my merge here, as I feel bad about adding confusing duplication to the source tree without further discussion. Sorry!

bmhatfield:
Please do feel encouraged to open a new PR or re-open this one if I've misunderstood. Apologies again.

PR #382, opened 2016-01-26
Add fluentd collector

csabiftw:
Added flutend collector that uses the fluentd plugin api.


src/collectors/fluentd/fluentd.py
```diff
@@ -0,0 +1,69 @@
...
+    host = localhost
+    port = 24220
+    [[[collect]]]
+        kinesis = buffer_queue_length, buffer_total_queued_size, retry_count
+```
+
+"""
+
+import diamond.collector
+import requests
```
bmhatfield: I am not sure, but I feel like we generally don't add dependencies on things like `requests` when the built-in libraries will do. It's a little less convenient for the author for sure, but it's *much* more convenient for the many future users of this collector. A quick grep of the entire diamond source tree seems to indicate that `import requests` appears in no other places, plausibly for this exact reason.

src/collectors/fluentd/fluentd.py
```diff
@@ -0,0 +1,69 @@
...
+        status = r.json()
+
+        result = self.parse_api_output(status)
+        for r in result:
+            self.publish(r[0], r[1])
+
+    def parse_api_output(self, status):
+        result = []
+        for p in status.get('plugins'):
+            if p['type'] in self.config['collect'].keys():
```
bmhatfield: It appears that this plugin will collect nothing without further configuration. Is there a way to automatically collect all the numeric plugin values by default, and make a whitelist optional instead of required?

csabiftw: I only use fluentd with the kinesis plugin, and I am not sure what structure other plugins use. It would be possible to do what you suggested, but i felt like stronger control over the collected values is more beneficial.

src/collectors/fluentd/fluentd.py
csabiftw: I have removed the dependency on requests in favor of urllib2 and json.

csabiftw:
The python 2.7 test seem to break for reasons unrelated to this PR.

bmhatfield:
Looks great, thanks!

PR #383, opened 2016-01-27
handlers: handler for app enlight metrics

ergo:


bmhatfield:
This overall seems good to me, but I'd love a double check from someone ( @jaingaurav ? )

ergo:
@bmhatfield  I think most of current handlers don't have tests, (probably because it would require exposing api keys to the public)? I did test the integration manually though.

shortdudey123:
needs doc file in `docs/handlers`

PR #386, opened 2016-01-29
Allow SNMP collector to address different communities

mbt-msw:
The SNMP collector uses the constant string 'agent' as the index for its CommunityData object.  This means multiple SNMP collectors will conflict, as second and subsequent collectors will use the same community name as the first collector, even if the diamond configuration specifies a different community.

Instead, include the community name in the index, so that each community has a unique index.


bmhatfield:
What's the effect of this change on existing setups?

krbaker:
This helped me out

josegonzalez:
Merged, thanks for the pull request!

PR #389, opened 2016-02-09
Fix invalid characters in kafka metric names

avishai-ish-shalom:


src/collectors/kafkastat/kafkastat.py
```diff
@@ -128,7 +128,8 @@ def query_mbean(self, objectname, key_prefix=None):
                         key_prefix = key_prefix + '.' + key
 
         metrics = {}
-
+        trans_table = string.maketrans(', ','__')
```
bmhatfield: Silly question, but why two `_` characters instead of 1?

bmhatfield: Ah, just read the maketrans doc more carefully. This all looks good to me.

bmhatfield:
Thanks!

src/collectors/kafkastat/kafkastat.py
jaingaurav: Seems this change broke travis tests:

https://travis-ci.org/python-diamond/Diamond/builds/108056178

PR #390, opened 2016-02-09
Revert "Fix invalid characters in kafka metric names"

bmhatfield:
Reverts python-diamond/Diamond#389
It broke tests.

cc @avishai-ish-shalom


bmhatfield:
Selfie due to broken tests.

PR #393, opened 2016-02-12
haproxy.py: Support unix socket collection

nyerup:
In addition to HTTP, HAProxy also exposes statistics through its
configurable admin UNIX domain socket, by use of the command 'show
stat'. This change enables support in the HAProxy collector for using
this for data collection.

To use this, set 'method' to 'unix' in the collector configuration, and
optionally set the socket path with the 'sock' option.

The default behaviour is left unchanged, as the 'method' option defaults
to 'http'.

### DIFF
--- a/src/collectors/haproxy/haproxy.py
+++ b/src/collectors/haproxy/haproxy.py
@@ -14,6 +14,7 @@ import re
 import urllib2
 import base64
 import csv
+import socket
 import diamond.collector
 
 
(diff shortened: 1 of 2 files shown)


nyerup:
For the record, I ran »make test« here, and got errors too. The errors were from the PortStatCollector however, and as far as I can tell not related to my patch. I see the same errors on head of master.

Please let me know if I introduced any errors.

jaingaurav:
@nyerup: The diff looks good. Could you please squash the commits?

Also, could you please add a test case/fixture for the unix domain socket path.

nyerup:
@jaingaurav Sure, I'll do that.

The current test mainly hooks in after the collector fetches metrics from the http interface, and the input data fetched from the unix is binarily identical. But I'll take a look, and try to add something meaningful.

nyerup:
I ended up implementing an actual mockup socket logic in the test case, to actually verify proper communications with the admin interface.

jaingaurav:
@nyerup: Awesome thanks for the work!

PR #394, opened 2016-02-12
Fix nfs space collection

johnseekins:
Since nfs servers can have multiple mount points, using the major/minor os.stat returns on a system with multiple nfs mounts means we only actually get the stats from one of them. This change fixes that behavior.

### DIFF
--- a/src/collectors/diskspace/diskspace.py
+++ b/src/collectors/diskspace/diskspace.py
@@ -41,6 +41,8 @@ class DiskSpaceCollector(diamond.collector.Collector):
                 "A list of regex patterns. Any filesystem" +
                 " matching any of these patterns will be excluded from disk" +
                 " space metrics collection",
+            'no_major_minor': "Don't rely on the major/minor numbers from an os.stat"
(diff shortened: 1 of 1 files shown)


src/collectors/diskspace/diskspace.py
```diff
@@ -156,8 +161,12 @@ def get_file_systems(self):
                                        mount_point)
                         continue
 
-                    if (major, minor) in result:
+                    if (major, minor) in result \
+                            and mount_point not in self.no_major_minor:
```
bmhatfield: The default for `self.no_major_minor` seems to be `'nfs'`, which is more like a type rather than a mount point itself. Am I misreading something here, or did you intend to use `fs_type` here instead?

johnseekins: As a matter of fact I *did* mean to use fs_type. This patch came across terribly!

Thanks.

bmhatfield:
Thanks again for this patch. Do you mind updating the existing tests to exercise the behavior change introduced in this PR? I think that will help catch any remaining subtle bugs :-)

PR #398, opened 2016-02-15
Processresources perf doc config

vilkaspilkas:
This improvement makes process.as_dict() fetch only the required fields rather than everything. It makes the execution much faster (milliseconds vs seconds on the system with 1000 running processes, monitoring ~20 of them). It also adds one more use case example and makes the fields to be monitored configurable.

### DIFF
--- a/src/collectors/processresources/processresources.py
+++ b/src/collectors/processresources/processresources.py
@@ -11,10 +11,14 @@ Example config file ProcessResourcesCollector.conf
 enabled=True
 unit=B
 cpu_interval=0.1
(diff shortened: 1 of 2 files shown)


jaingaurav:
@vilkaspilkas: Thanks so much for this fix. I've been wanting to do something like this for the longest time. I'm very happy that you were able to do this. Could you please squash the commits so that I can merge it?

vilkaspilkas:
@jaingaurav: Hey, is it ok now?

jaingaurav:
Yes thanks a lot. I look forward to observing the improvements.

PR #402, opened 2016-02-20
Allow scheme for rabbitmq url

beedub:


bmhatfield:
This looks good to me, passes the checks, and seems backwards compatible. Thanks!

PR #403, opened 2016-02-20
Revert "Add docker stats collector"

bmhatfield:
Reverts python-diamond/Diamond#381

Dang, thought I'd do a double check and noticed that there was already a docker collector with *very* high overlap. Reverting until we are able to resolve that.


bmhatfield:
There's a weird cython error on one of the tests, but everything else passed and I don't see how immediately reverting this PR could cause breakage here.

csabiftw:
I added this collector instead of the original docker collector because:
- this uses stats API, which is not a significant thing, but standard.
- this collector can also collect CPU info, which is not possible with the currently available one, and I need this functionality

In the light of this, I am reopening this PR, and we can continue the discussion there.

PR #404, opened 2016-02-22
Add docker stats collector

csabiftw:
I added this collector instead of the original docker collector because:

- this uses stats API, which is not a significant thing, but standard.
- this collector can also collect CPU info, which is not possible with the currently available one, and I need this functionality

Original discussion here: https://github.com/python-diamond/Diamond/pull/403


bmhatfield:
Thanks for reopening this PR! I think the code here is great, the only real question I have is about whether or not we should have two competing docker collectors.

We have a few options:
- Remove the other collector, breaking folks but standardizing on this one.
- Merge the collectors together, with options, adding some complexity.
- Keep both collectors, potentially confusing users
- Don't accept this one, missing out on a potentially more standard approach to gathering this data.

From a user perspective, I think having two distinct but very similar collectors is really confusing. I am not sure of the best way forwards otherwise.

jfarrell:
+1 to unifying what the other docker collector did into the newer api based version and only having one

csabiftw:
Based on the discussion, this should be a dropin replacement for the old collector, with some additional metrics.

An example output:
```
docker.images_count 4 1456163431
docker.images_dangling_count 0 1456163431
docker.container-name.RSS 5410816 1456163431
docker.container-name.cache 9355264 1456163431
docker.container-name.cpu.kernelmode 20000000 1456163431
docker.container-name.cpu.system 128529160000000 1456163431
docker.container-name.cpu.total 117645054 1456163431
docker.container-name.cpu.usermode 60000000 1456163431
docker.container-name.pagein_count 4632 1456163431
docker.container-name.pageout_count 2049 1456163431
docker.container-name.swap 0 1456163431
```

bmhatfield:
This is amazing! Thank you!

PR #406, opened 2016-02-22
Changed stdout flag to add loggers from conf file and honor root log level

csabiftw:
This PR changes the behaviour of the  --log-stdout flag.

Previously, this flag disregarded any loggers defined in the config file, and created a new logger with DEBUG log level. 

This means, that if you use systemd, there's no way to set your log level to anything but DEBUG using the default rpm package: https://github.com/python-diamond/Diamond/blob/master/rpm/systemd/diamond.service#L5

With this PR, loggers defined in the conf file still get initialised, and the stdout loggers level will be the same as the root loggers.

### DIFF
--- a/src/diamond/utils/log.py
+++ b/src/diamond/utils/log.py
@@ -35,27 +35,32 @@ class DebugFormatter(logging.Formatter):
 def setup_logging(configfile, stdout=False):
     log = logging.getLogger('diamond')
 
-    if stdout:
-        log.setLevel(logging.DEBUG)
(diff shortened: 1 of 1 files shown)


joel-airspring:
This (or something like this patch or even just removing --log-stdout from diamond.service) would be helpful.  I just burned a few hours tweaking the log levels in my config files, etc., trying to figure out why I was getting so much verbose DEBUG output even though the configuration file had log level INFO.

Turns out the configuration file log settings are ignored with --log-stdout, which is the default setting defined for systemd services in diamond.service

PR #408, opened 2016-02-23
Fix globbing issue when metrics list is length 1

atoz3q:
In the HBase and Hadoop collectors, if the `metrics` list of globbing patterns contains only one element, the loop in `collect()` will iterate over the characters in the string and glob over the individual characters, instead of globbing over the string as a whole. A check for if the `metrics` value is a `list` or not determines whether there is only one value provided or not.

### DIFF
--- a/src/collectors/hadoop/hadoop.py
+++ b/src/collectors/hadoop/hadoop.py
@@ -44,7 +44,11 @@ class HadoopCollector(diamond.collector.Collector):
         return config
 
     def collect(self):
-        for pattern in self.config['metrics']:
+        metrics = self.config['metrics']
+        if not isinstance(metrics, list):
+            metrics = [str(metrics)]
+
+        for pattern in metrics:
             for filename in glob.glob(pattern):
                 self.collect_from(filename)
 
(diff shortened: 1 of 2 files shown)


src/collectors/hadoop/hadoop.py
```diff
@@ -44,7 +44,12 @@ def get_default_config(self):
         return config
 
     def collect(self):
-        for pattern in self.config['metrics']:
+        if isinstance(self.config['metrics'], list):
```
jaingaurav: I think a better approach would be to change the code to always take a list of patterns. Something like this:
```
metrics = self.config['metrics']
if not isinstance(metrics, list):
    metrics = [str(metrics)]

for metric in metrics:
    for filename in glob.glob(metric):
        self.collect_from(filename)
```

qardanaz: Yes, you are right -- this is the better way to do it. I've made the changes and tested them and it works just fine as far as I can tell.

jaingaurav:
Thanks, looks great now. Only minor request is to squash the commits and I'll then merge it in.

qardanaz:
Thanks! I've squashed the three commits so we should be set to merge.

PR #412, opened 2016-03-01
socket.settimeout needs a float

cannium:
see https://docs.python.org/2/library/socket.html#socket.socket.settimeout


jaingaurav:
Thanks

PR #413, opened 2016-03-02
Respawn diamond service on crash

iurii-polishchuk:


iurii-polishchuk:
ci build failed due to "sh: 1: cython: not found". What's wrong?

csabiftw:
This probably has nothing to do with your PR, as it has happened to me too. Try triggering the build again and see if it passes.

iurii-polishchuk:
yes, it passed. thanks

redref:
Sorry to say it, but I don't clearly see purpose of this "respawn". If something goes wrong, daemon may be respawned indefinitly causing other issues (unless I missed some protection).

If one collector/handler cause diamond to crash on some circonstances, it would be quite better to find and fix the problem.

iurii-polishchuk:
@redref I found that on our prod systems diamond is killed periodically (once a day or two):
`[Tue Mar  1 16:03:47 2016] init: diamond main process (998) killed by KILL signal [Tue Mar  1 16:03:47 2016] init: diamond main process (9698) terminated with status 1`
Didn't manage to find the reason. That looks like OOM killer, but instance is not under high load, there's 1GB of free RAM, load avg is 0.2.
In diamond log there's nothing suspicious.
So I added 'respawn' to upstart config to keep diamond running.
Does it make sense?

redref:
You got no log from oom killer ? Maybe collector/handler get stuck and reached 1GB data. Can you compare last datapoint with kill timestamp ?

Got huge memory use myself when having network issues.

iurii-polishchuk:
Nope, OOM killer was just assumption. When it kills the process, there's 'out of memory' message in dmesg. So it's something else.

Yes, last datapoint stamp is less than 1 minute before crash.

redref:
You meant "there's no 'out of memory' message", right ?

No clue there. If it's SIGKILL, it looks like OOM Killer but is'nt. If there is no log, nor messages, it's kinda a dead end.

iurii-polishchuk:
Found that diamond was killed at the same time on two instances. The time of crash is exact when we have deployment. Diamond was stopped by ansible, but not started for some reason.

jaingaurav:
@iurii-polishchuk: Could you paste the ansible rule that caused diamond to stop. Perhaps we can assist there? You might find out exactly what ansible is doing from syslog or auth.log

iurii-polishchuk:
Seems the reason is /etc/init/diamond.conf. Removing SysV config fixes the issue.

PR #416, opened 2016-03-04
added option to replace slashes in rabbitmq queue or virtualhost names

merlijntishauser:
RabbitMQ has a default vhost named "/" in certain setups.
This causes a problem with forwarding metrics to hostedGraphite. With grafana in hostedGraphite you can't select the metrics with / as vhost. This is solved by replacing / in the queue or vhost name by a custom string.


jaingaurav:
Looks great, thanks

PR #417, opened 2016-03-04
add a collector for Eventstore Projections

merlijntishauser:
In this PR I have added a collector for [Eventstore](https://geteventstore.com/) projections. It has only be tested against a single node version of eventstore (not clustered). We use it already in production and does it's job nicely in combination with hostedGraphite. Note that $ in projection names are replaced by a _ to avoid problems in Grafana.

The collector is based on the HTTPJSON collector, which could not handle the slightly more complex json output from the eventstore api.


src/collectors/eventstoreprojections/eventstoreprojections.py
```diff
@@ -0,0 +1,100 @@
...
+        })
+        return config_help
+
+    def get_default_config(self):
+        default_config = super(
+            EventstoreProjectionsCollector, self).get_default_config(
+        )
+        default_config.update({
+            'path': 'eventstore',
+            'url': 'http://hostname:2113/projections/all-non-transient',
```
jaingaurav: could you change this from `hostname` to `localhost`? At least then it is somewhat of a legitimate default config.

src/collectors/eventstoreprojections/eventstoreprojections.py
```diff
@@ -0,0 +1,100 @@
...
+class EventstoreProjectionsCollector(diamond.collector.Collector):
+
+    def get_default_config_help(self):
+        config_help = super(
+            EventstoreProjectionsCollector, self).get_default_config_help(
+        )
+        config_help.update({
+            'url': 'Full URL',
+            'headers': 'Header variable if needed. '
+            'Will be added to every request',
```
jaingaurav: Could you please keep the lines of the description at the same indent level? This might require wrapping it in parenthesis.

src/collectors/eventstoreprojections/eventstoreprojections.py
```diff
@@ -0,0 +1,100 @@
...
+            elif isinstance(value, basestring):
+                if value == "Running":
+                    value = 1
+                    yield ("%s.%s" % (prefix, key), value)
+                elif value == "Stopped":
+                    value = 0
+                    yield ("%s.%s" % (prefix, key), value)
+                else:
+                    # very chatty, uncomment if you want to see the output
+                    # self.log.debug("ignoring string value = %s", value)
```
jaingaurav: Rather than keeping this uncommented, please add a `debug` config option

jaingaurav:
Gave some very minor comments mainly around the config options. Please make the edits and I'll merge this in.

merlijntishauser:
Added a commit where the comments are addressed. Thanks for the feedback!

src/collectors/eventstoreprojections/eventstoreprojections.py
```diff
@@ -0,0 +1,110 @@
...
+        })
+        return config_help
+
+    def get_default_config(self):
+        default_config = super(
+            EventstoreProjectionsCollector, self).get_default_config(
+        )
+        default_config.update({
+            'path': "eventstore",
+            'url': 'http://localhost',
```
jaingaurav: So I still feel like this option should be `hostname` instead of url. Might be best to break it up into 2 options `protocol` and `hostname`

src/collectors/eventstoreprojections/eventstoreprojections.py
```diff
@@ -0,0 +1,110 @@
...
+            self.log.error("Can't open url %s. %s", url, e)
+        else:
+            content = resp.read()
+            try:
+                json_dict = json.loads(content)
+                projections = json_dict['projections']
+
+                data = {}
+                for projection in projections:
+                    name = projection["name"].replace("$", "_")
```
jaingaurav: Does it make sense to replace `$` with `_` or simply just delete the `$`. The delete option would make nicer metric names, however I don't know if there could be a name collision.

merlijntishauser:
Split up the url into 2 new config options as proposed, I misunderstood the initial comment. 
Just deleting the $ would almost for sure give name collisions. As a solution, I introduced a config option so that you, if you really want, could skip replacing the $ or replace it by a string of your choice.
As done in the rabbitMQ collector for dots and/or slashes.

Unfortunately, travis fails on the updated PR for python 2.6 with the message: "The command "pip install -r .travis.requirements.txt" failed and exited with 1 during ." It seems very unlikely this is caused by my changes in the last 2 commits. So it should be safe to merge ;)

jaingaurav:
Thanks a lot, I re-ran travis tests and it looks good now. The failure you mentioned is a known flakiness.

PR #420, opened 2016-03-08
Update influxdbHandler.py

Coolknight:
I took from github a version of influxdbHandler.py which differences between 0.8 and 0.9 and added support for tags and series.

original series example: servers.ubuntu.iostat.sda1.writes_merged
with this code: 
serie: iostat
tags: host=ubuntu,iostat=sda1
metric: writes_merged

Hope it helps.

### DIFF
--- a/src/diamond/handler/influxdbHandler.py
+++ b/src/diamond/handler/influxdbHandler.py
@@ -246,12 +246,12 @@ class InfluxdbHandler(Handler):
             if self.influxdb_version == '0.8':
                 # Use legacy client for InfluxDB 0.8
                 self.influx = InfluxDB08Client(self.hostname, self.port,
-                    self.username, self.password,
-                    self.database, self.ssl)
+                                               self.username, self.password,
+                                               self.database, self.ssl)
             else:
                 self.influx = InfluxDBClient(self.hostname, self.port,
-                    self.username, self.password,
-                    self.database, self.ssl)
+                                             self.username, self.password,
+                                             self.database, self.ssl)
             # Log
             self.log.debug("InfluxdbHandler: Established connection to "
                            "%s:%d/%s.",
--- a/src/diamond/handler/influxdbHandler.py
+++ b/src/diamond/handler/influxdbHandler.py
@@ -10,12 +10,16 @@ v1.2 : added a timer to delay influxdb writing in case of failure
        this whill avoid the 100% cpu loop when influx in not responding
        Sebastien Prune THOMAS - prune@lecentre.net
 
-#### Dependencies
- * [influxdb](https://github.com/influxdb/influxdb-python)
+- Dependency:
+    - influxdb client (pip install influxdb)
+      you need version > 0.1.6 for HTTPS (not yet released)
 
+- enable it in `diamond.conf` :
 
-#### Configuration
-```
+handlers = diamond.handler.influxdbHandler.InfluxdbHandler
+
+- add config to `diamond.conf` :
+
 [[InfluxdbHandler]]
 hostname = localhost
 port = 8086 #8084 for HTTPS
@@ -25,9 +29,10 @@ username = root
 password = root
 database = graphite
 time_precision = s
-```
+influxdb_version = 0.9
 """
 
+from six import integer_types
 import time
 from Handler import Handler
 
@@ -36,24 +41,22 @@ try:
 except ImportError:
     InfluxDBClient = None
 
+try:
+    from influxdb.influxdb08 import InfluxDBClient as InfluxDB08Client
+except ImportError:
+    InfluxDB08Client = None
+
 
 class InfluxdbHandler(Handler):
     """
     Sending data to Influxdb using batched format
     """
-
     def __init__(self, config=None):
         """
         Create a new instance of the InfluxdbeHandler
         """
         # Initialize Handler
         Handler.__init__(self, config)
-
-        if not InfluxDBClient:
-            self.log.error('influxdb.client.InfluxDBClient import failed. '
-                           'Handler disabled')
-            self.enabled = False
-            return
 
         # Initialize Options
         if self.config['ssl'] == "True":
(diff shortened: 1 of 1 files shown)


bmhatfield:
Hi,

A few points of feedback. First, I don't think that you can generically just split metric paths into `key, value` pairs. I only briefly reviewed the PR at this time, but that appeared to be the approach. This was a strong and common point of feedback for the InfluxDB team when they changed this functionality.

See here: https://github.com/influxdata/influxdb/issues/2102 and here https://github.com/influxdata/influxdb/pull/3125 for more information and discussion.

Second, your code only appears to deal with 0.8 and 0.9 - but 0.10 is the current stable version, as far as I am aware.

Third, your PR is currently failing tests because of pep8 issues:

```
$ pep8 --config=.pep8 src bin/diamond bin/diamond-setup build_doc.py setup.py test.py
src/diamond/handler/influxdbHandler.py:84:81: E501 line too long (86 > 80 characters)
src/diamond/handler/influxdbHandler.py:85:29: E127 continuation line over-indented for visual indent
src/diamond/handler/influxdbHandler.py:196:81: E501 line too long (82 > 80 characters)
src/diamond/handler/influxdbHandler.py:205:81: E501 line too long (90 > 80 characters)
src/diamond/handler/influxdbHandler.py:206:81: E501 line too long (88 > 80 characters)
src/diamond/handler/influxdbHandler.py:207:81: E501 line too long (94 > 80 characters)
src/diamond/handler/influxdbHandler.py:247:41: E128 continuation line under-indented for visual indent
src/diamond/handler/influxdbHandler.py:248:41: E128 continuation line under-indented for visual indent
src/diamond/handler/influxdbHandler.py:251:42: E128 continuation line under-indented for visual indent
src/diamond/handler/influxdbHandler.py:252:42: E128 continuation line under-indented for visual indent
The command "pep8 --config=.pep8 src bin/diamond bin/diamond-setup build_doc.py setup.py test.py" exited with 1.
```

I'll also add some feedback on the PR itself.

src/diamond/handler/influxdbHandler.py
```diff
@@ -159,39 +176,63 @@ def _send(self):
...
+                    if self.influxdb_version == "0.8":
+                        for path in self.batch:
+                            metrics.append({
+                                "points": self.batch[path],
+                                "name": path,
+                                "columns": ["time", "value"]})
+                    elif self.influxdb_version == "0.9":
+                        for path in self.batch:
+                            # split path to use influx syntax for tags and tables.
+                            tokenized = path.split('.')
```
bmhatfield: Instead of accessing this with indexed keys, ie `tokenized[2]`, I think it would be good to assign them to variables that more clearly explain what you believe the field holds.

src/diamond/handler/influxdbHandler.py
```diff
@@ -159,39 +176,63 @@ def _send(self):
...
+                            table = tokenized[2]
+                            tags = {'host': tokenized[1]}
```
bmhatfield: I don't think that you know this is true. There are top-level configuration options for this in Diamond. I would expect any approach here to take them into account to determine what the `prefix` is, where the `host` field lives (if it does at all... in our case, for example, we have some metrics that do not have a host-specific field at all), etc.

src/diamond/handler/influxdbHandler.py
```diff
@@ -159,39 +176,63 @@ def _send(self):
...
+                else:
+                    # build metrics data
+                    metrics = []
+                    if self.influxdb_version == "0.8":
+                        for path in self.batch:
+                            metrics.append({
+                                "points": self.batch[path],
+                                "name": path,
+                                "columns": ["time", "value"]})
+                    elif self.influxdb_version == "0.9":
```
bmhatfield: Should this be 0.9 or greater?

src/diamond/handler/influxdbHandler.py
```diff
@@ -159,39 +176,63 @@ def _send(self):
...
+                        for path in self.batch:
+                            # split path to use influx syntax for tags and tables.
+                            tokenized = path.split('.')
+                            table = tokenized[2]
+                            tags = {'host': tokenized[1]}
+                            if len(tokenized) == 4:
+                                metricname = tokenized[3]
+                            elif len(tokenized) == 5:
```
bmhatfield: What happens if my metric path has more than 5 components?

src/diamond/handler/influxdbHandler.py
Coolknight: Yes, currently i am using 0.10.1 and this works for me so it should be 0.9 or greater but my coding skills are null and i didn't know how to compare strings. U_U'

src/diamond/handler/influxdbHandler.py
Coolknight: I think i am doing exactly what you say:

table = tokenized[2] <- on the third token there is the table name
tags = {'host': tokenized[1]} <- on the second token there is the value of the tag host.

src/diamond/handler/influxdbHandler.py
bmhatfield: Yep - version math can be tricky. Perhaps try something like this: http://stackoverflow.com/a/6972866 (other suggestions in that thread might be helpful as well).

src/diamond/handler/influxdbHandler.py
Coolknight: Correct, this approach should only work where you havent configured the prefix. Correct aproach i think it should cut the prefix from the path and split what is left. But again i don't have the necessary skill to do it.

src/diamond/handler/influxdbHandler.py
Coolknight: All metrics i've seen are 4 or 5 components long. If i find a different metric length i will update the handler.

Coolknight:
Thanks a lot for your feedback bmhatfield.

From your first point, yes, i split metric paths into key value pairs. Maybe it is not the perfect approach but indeed it is better than 1 table each metric path.

From your second point, I believe there was a huge change from 0.8 to 0.9 and there is no retrocompatibility. But there is compatibility between 0.9 and 0.10, so i didn't bother on learning how to compare those strings. Because that's the main issue on my side, i was learning python as i was reading code here.

From your third point, i will change those lines but i think they are more warnings than errors because handler is working.

I am sorry but i am not a coder willing to help on the project, im just a user who took a roundabout to make things work and i wanted to share.

shortdudey123:

PR #430, opened 2016-03-12
Updated Dropwizard metrics format

jlordiales:
The current Dropwizard collector uses a pretty old format for the expected metrics JSON that DW returns. This is completely broken with the latest version of DW.

This PR fixes that and works for all existing metrics (as opposed to just JVM metrics as per the current behaviour).


shortdudey123:
@jlordiales appears that this breaks backwards compatibility, that correct?

jlordiales:
@shortdudey123 Yeah. On the other hand, every Dropwizard version that has come out in the last 4 years (at least) has this new format. So I assume most, if not all, using DW can't use the current collector anyway.

PR #432, opened 2016-03-14
Collect pgfault and pgmajfault from vmstat

shortdudey123:
Collects the minor and major page faults
See http://www.linuxinsight.com/proc_vmstat.html for details on `/proc/vmstat`


jaingaurav:
Thanks, merged

PR #442, opened 2016-03-24
RabbitMQCollector: Add a query_individual_queues config option

jwheare:
For vhosts with a large number of queues, fetching all of them at once can time out.
If queues are specified, this option will ensure that they are queried individually

### DIFF
--- a/src/collectors/rabbitmq/test/testrabbitmq.py
+++ b/src/collectors/rabbitmq/test/testrabbitmq.py
@@ -309,7 +309,7 @@ class TestRabbitMQCollector(CollectorTestCase):
         client.get_vhost_names.return_value = ['vhost1', 'vhost2']
 
         self.collector.collect()
-        
+
         client.get_queues.assert_not_called()
         client.get_nodes.assert_called_once_with()
         client.get_node.assert_called_once_with('rabbit@localhost')
@@ -320,7 +320,7 @@ class TestRabbitMQCollector(CollectorTestCase):
             call('vhost2', 'queue3'),
             call('vhost2', 'queue4'),
         ], any_order=True)
-        
+
         metrics = {
             'queues.queue1.key': 1,
             'queues.queue2.key': 2,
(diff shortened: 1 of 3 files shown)


jwheare:
This has issues related to legacy vhostless configs. Closing while I fix.

jwheare:
OK, fixed this for legacy. vhost names are now always expanded and specified in api calls.

Let me know if you'd like this squashed before merge.

shortdudey123:
If the root cause of the change is the api calls timing out, what is the performance difference between just increasing the timeout vs this PR?

jwheare:
The root issue is we have a huge number of queues and only need to query some of them. So rather than wait for a very large response to arrive and then postfilter, we only query the subset we need. Increasing the timeout is not an equivalent solution.

src/collectors/rabbitmq/test/testrabbitmq.py
```diff
@@ -237,6 +243,219 @@ def test_opt_should_replace_slashes(self, publish_mock, client_mock):
...
+            'health.sockets_total': 2,
+            'health.disk_free_limit': 1,
+            'health.disk_free': 1,
+            'health.proc_used': 1,
+            'health.proc_total': 1,
+            'cluster.partitions': 0,
+            'cluster.nodes': 3
+        }
+
+        self.setDocExample(collector=self.collector.__class__.__name__,
```
shortdudey123: remove `self.setDocExample` here (must only be called once per test file)

src/collectors/rabbitmq/test/testrabbitmq.py
```diff
@@ -237,6 +243,219 @@ def test_opt_should_replace_slashes(self, publish_mock, client_mock):
...
```
shortdudey123: remove `self.setDocExample` here (must only be called once per test file)

shortdudey123:
Gotcha, can you fix the 2 comments i left?

jwheare:
Done. I just copied the other examples in that file. It's currently called 3 times in that file fwiw.

PR #451, opened 2016-04-13
Add `_normalized` metrics for loadaverage

shortdudey123:
Normalized is the load average divided by the number of CPU's
A normalized load average of 1 = all CPU's at 100% usage


jaingaurav:
Very useful thanks!

PR #453, opened 2016-04-15
Add ntp collector

shortdudey123:
Collect out of band stats from ntp

Uses output from ntpdate:

```
$ ntpdate -q pool.ntp.org
server 12.34.56.1, stratum 2, offset -0.000277, delay 0.02878
server 12.34.56.2, stratum 1, offset -0.000128, delay 0.02896
server 12.34.56.3, stratum 2, offset 0.000613, delay 0.02870
server 12.34.56.4, stratum 2, offset -0.000351, delay 0.02864
31 Apr 12:00:00 ntpdate[12]: adjust time server 12.34.56.2 offset -0.000128 sec
$

### DIFF
--- a/docs/collectors/NtpCollector.md
+++ b/docs/collectors/NtpCollector.md
@@ -1,0 +1,49 @@
+<!--This file was generated from the python source
+Please edit the source to make changes
+-->
+NtpCollector
+=====
+
+Collect out of band stats from ntp
+
+Uses output from ntpdate:
+
+```
+$ ntpdate -q pool.ntp.org
+server 12.34.56.1, stratum 2, offset -0.000277, delay 0.02878
+server 12.34.56.2, stratum 1, offset -0.000128, delay 0.02896
+server 12.34.56.3, stratum 2, offset 0.000613, delay 0.02870
+server 12.34.56.4, stratum 2, offset -0.000351, delay 0.02864
+31 Apr 12:00:00 ntpdate[12]: adjust time server 12.34.56.2 offset -0.000128 sec
+$
+```
+
+#### Dependencies
+
+    * /usr/sbin/ntpdate
+    * subprocess
+
+
+#### Options
+
+Setting | Default | Description | Type
+--------|---------|-------------|-----
+bin | /usr/sbin/ntpdate | Path to ntpdate binary | str
+byte_unit | byte | Default numeric output(s) | str
+enabled | False | Enable collecting these metrics | bool
+measure_collector_time | False | Collect the collector run time in ms | bool
+metrics_blacklist | None | Regex to match metrics to block. Mutually exclusive with metrics_whitelist | NoneType
+metrics_whitelist | None | Regex to match metrics to transmit. Mutually exclusive with metrics_blacklist | NoneType
+ntp_pool | pool.ntp.org | NTP Pool address | str
+precision | 0 | Number of decimal places to report to | int
+sudo_cmd | /usr/bin/sudo | Path to sudo | str
+time_scale | milliseconds | Time unit to report offset in | str
+use_sudo | False | Use sudo? | bool
+
+#### Example Output
+
+```
+servers.hostname.ntp.offset.milliseconds 0
+servers.hostname.ntp.server.count 4
+```
+
--- a/src/collectors/ntp/ntp.py
+++ b/src/collectors/ntp/ntp.py
@@ -1,0 +1,99 @@
+# coding=utf-8
+
+"""
+Collect out of band stats from ntp
+
+Uses output from ntpdate:
+
+```
+$ ntpdate -q pool.ntp.org
+server 12.34.56.1, stratum 2, offset -0.000277, delay 0.02878
+server 12.34.56.2, stratum 1, offset -0.000128, delay 0.02896
+server 12.34.56.3, stratum 2, offset 0.000613, delay 0.02870
+server 12.34.56.4, stratum 2, offset -0.000351, delay 0.02864
+31 Apr 12:00:00 ntpdate[12]: adjust time server 12.34.56.2 offset -0.000128 sec
+$
+```
+
+#### Dependencies
+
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by shortdudey123** [comment]
Hey @jaingaurav 
Can you look at this PR? as well as a couple of my other ones? Thanks!!!

File: src/collectors/ntp/ntp.py [hunk 1]
```diff
@@ -0,0 +1,85 @@
...
+            # Sample of line: 31 Apr 12:00:00 ntpdate[12345]: adjust time \
+            #   server 123.456.789.2 offset -0.000123 sec
+            if len(parts) != 11:
+                continue
+
+            # offset is in seconds, convert is to nanoseconds and miliseconds
+            offset_in_s = float(parts[9])
+            offset_in_ms = offset_in_s * 1000
+            offset_in_us = offset_in_s * 1000000
+            data['offset_s'] = {'val': offset_in_s, 'precision': 6}
```
jaingaurav: So rather than having all these offset stats, how about you have the unit and precision in the config as an option and then use the unit conversion utils?

shortdudey123: ah, do you have a good example of that?

src/collectors/ntp/ntp.py
```diff
@@ -0,0 +1,85 @@
...
+    def get_ntpdate_stats(self):
+        output = self.run_command(['-q', self.config['ntp_pool']])
+
+        data = {}
+        data['server_count'] = {'val': 0, 'precision': 0}
+
+        for line in output[0].splitlines():
+            # Only care about best choice not all servers
+            if line.startswith('server'):
+                data['server_count']['val'] += 1
```
jaingaurav: I don't really get the point of adding `server_count` here. It seems that this is more of an index than a count.

src/collectors/ntp/ntp.py
jaingaurav: https://github.com/python-diamond/Diamond/blob/master/src/collectors/uptime/uptime.py#L44

src/collectors/ntp/ntp.py
```diff
@@ -0,0 +1,85 @@
...
+                continue
+
+            parts = line.split()
+
+            # Make sure we have the correct output
+            # Sample of line: 31 Apr 12:00:00 ntpdate[12345]: adjust time \
+            #   server 123.456.789.2 offset -0.000123 sec
+            if len(parts) != 11:
+                continue
```
jaingaurav: Could you log an error here?

docs/collectors/NtpCollector.md
```diff
@@ -0,0 +1,49 @@
...
+metrics_blacklist | None | Regex to match metrics to block. Mutually exclusive with metrics_whitelist | NoneType
+metrics_whitelist | None | Regex to match metrics to transmit. Mutually exclusive with metrics_blacklist | NoneType
+ntp_pool | pool.ntp.org | NTP Pool address | str
+sudo_cmd | /usr/bin/sudo | Path to sudo | str
+use_sudo | False | Use sudo? | bool
+
+#### Example Output
+
+```
+servers.hostname.ntp.offset_ms -0.128
```
jaingaurav: Was this generated The `server.hostname` prefix seems incorrect

shortdudey123: yes this was generated

src/collectors/ntp/ntp.py
shortdudey123: 👍

shortdudey123:
@jaingaurav made changes per feedback
Will squash this down when you are good with the PR

src/collectors/ntp/ntp.py
```diff
@@ -0,0 +1,95 @@
...
+            if len(parts) != 11:
+                self.log.error('NtpCollector: Output of ntpdate was %s words '
+                               'long but was expected to be 11' % len(parts))
+                self.log.debug('NtpCollector: ntpdate output was %s' % parts)
+                continue
+
+            # offset is in seconds, convert is to nanoseconds and miliseconds
+            offset_in_s = float(parts[9])
+
+            data['offset_s'] = {'val': offset_in_s,
```
jaingaurav: So maybe I wasn't quite clear, but my suggestion was why not use a configuration option such as `units` and then only have a single `offset` metric published, or am I missing something here? Also you may want to expose a `precision` option too then.

shortdudey123: I was looking to have atleast msec and usec exposed but i just exposed sec as well

Is it safe to make the assumption that people are ok with their metric changing scale if they change the config option on a single metric namespace? (i.e. having sec and msec on the same metric name)

jaingaurav: So why do you want both msec and usec exposed?

Regarding your latter point, you can stick with the style that the uptime collector has or use a suffix like you've currently done.

shortdudey123:
> So why do you want both msec and usec exposed?

Actually, i looked a bit more and in reality i only need one

>  you can stick with the style that the uptime collector has or use a suffix like you've currently done.

I looked more at the uptime collector, and it sends the metrics in a different namespace depending on the scale you use. I missed that so its probably better if i follow that style :)

src/collectors/ntp/ntp.py
shortdudey123: >So why do you want both msec and usec exposed?

Actually, i looked a bit more and in reality i only need one

>you can stick with the style that the uptime collector has or use a suffix like you've currently done.

I looked more at the uptime collector, and it sends the metrics in a different namespace depending on the scale you use. I missed that so its probably better if i follow that style :)

shortdudey123: @jaingaurav updated

shortdudey123:
@jaingaurav changed `server_count` to `server.count` and squashed :)

jaingaurav:
Thanks a lot

Issue #225, opened 2015-07-17
Kafka Collector error with urllib2


maauso:
Hi, 
I try connect kafka collection with kafka 0.8.2. 

[2015-07-17 11:44:56,298] [MainThread] ''
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/diamond/collector.py", line 472, in _run
    self.collect()
  File "/usr/share/diamond/collectors/kafkastat/kafkastat.py", line 163, in collect
    match = self.get_mbeans(pattern)
  File "/usr/share/diamond/collectors/kafkastat/kafkastat.py", line 84, in get_mbeans
    mbeans = self._get('/serverbydomain', query_args)
  File "/usr/share/diamond/collectors/kafkastat/kafkastat.py", line 70, in _get
    response = urllib2.urlopen(url)
  File "/usr/lib64/python2.7/urllib2.py", line 127, in urlopen
    return _opener.open(url, data, timeout)
  File "/usr/lib64/python2.7/urllib2.py", line 404, in open
    response = self._open(req, data)
  File "/usr/lib64/python2.7/urllib2.py", line 422, in _open
    '_open', req)
  File "/usr/lib64/python2.7/urllib2.py", line 382, in _call_chain
    result = func(*args)
  File "/usr/lib64/python2.7/urllib2.py", line 1216, in http_open
    return self.do_open(httplib.HTTPConnection, req)
  File "/usr/lib64/python2.7/urllib2.py", line 1189, in do_open
    r = h.getresponse(buffering=True)
  File "/usr/lib64/python2.7/httplib.py", line 1045, in getresponse
    response.begin()
  File "/usr/lib64/python2.7/httplib.py", line 409, in begin
    version, status, reason = self._read_status()
  File "/usr/lib64/python2.7/httplib.py", line 373, in _read_status
    raise BadStatusLine(line)
BadStatusLine: ''

Any idea?

iurii-polishchuk:
I've got the same issue. It seems some kafka http reporter is required to expose jmx via http.

iurii-polishchuk:
Also actual documentation for KafkaCollector is highly appreciated.

iurii-polishchuk:
[...]

PR #454, opened 2016-04-17
Refactor KafkaCollector metric key generation. This helps to get predictable (without any invalid characters) metric keys

shamil:
The current implementation of metric key generation doesn't take into account invalid characters and some other cases where last keys can have dots (.) in the object name.

**For example, in old implementation, object-name**

`kafka.server:type=FetcherLagMetrics,name=ConsumerLag,clientId=ReplicaFetcherThread-0-60374353,topic=test,partition=0`

Will generate metric key

`FetcherLagMetrics.ConsumerLag,clientId.ConsumerLag,clientId.ConsumerLag,clientId.ConsumerLag,clientId`

which pretty useless. First, it contains commas which unsupported by Graphite or a like, second it has repeated metric names.

The new implementation will generate the following metric key

`kafka.server.FetcherLagMetrics.name_ConsumerLag.clientId_ReplicaFetcherThread-0-60374353.topic_test.partition_0`

---

**Another example for object-name**

`kafka.server:type=FetcherStats,name=RequestsPerSec,clientId=ReplicaFetcherThread-0-748817341,brokerHost=10.121.5.4,brokerPort=6667`

In old will be

`FetcherStats.RequestsPerSec,clientId.RequestsPerSec,clientId.RequestsPerSec,clientId.RequestsPerSec,clientId`

In new will be

`kafka.server.FetcherStats.name_RequestsPerSec.clientId_ReplicaFetcherThread-0-748817341.brokerHost_10_121_5_4.brokerPort_6667`

---

Just to be sure that I'm on the same page. I'm using Kafka 0.8.x. Which could mean that the current KafkaCollector doesn't support 0.8.x versions, hence the problematic metrics. I don't have older Kafka to test against, so if this the case I can submit the collector as V2 or something, up to you.

Also, I've updated the tests cases to support the new output format. Please review and suggest any changes, I'm open to every desired change just to make this into mainline, cause we rely on this collector in our Monitoring solution...

Thanks

### DIFF
--- a/src/collectors/kafkastat/kafkastat.py
+++ b/src/collectors/kafkastat/kafkastat.py
@@ -109,10 +109,10 @@ class KafkaCollector(diamond.collector.Collector):
 
         # extract prefix & keys from the obkect name
         # for example:
-        #  if objectname = kafka.network:type=RequestMetrics,name=RequestQueueTimeMs,request=OffsetCommit
-        #  prefix will be: kafka.network
-        #  keys   will be: ['type=RequestMetrics', 'name=RequestQueueTimeMs', 'request=OffsetCommit']
-        (prefix, keys)= objectname.split(':')
+        #  objectname: kafka.log:type=LogFlushStats,name=LogFlushRateAndTimeMs
+        #  prefix will be: kafka.log
+        #  keys will be: ['type=LogFlushStats', 'name=LogFlushRateAndTimeMs']
+        (prefix, keys) = objectname.split(':')
         keys = keys.split(',')
 
         if key_prefix is None:


src/collectors/kafkastat/test/testkafka.py
```diff
@@ -108,10 +108,10 @@ def test_query_mbean(self, get_mock):
         get_mock.return_value = self._get_xml_fixture('mbean.xml')
 
         expected_metrics = {
-            'kafka.logs.mytopic-1.CurrentOffset': long('213500615'),
-            'kafka.logs.mytopic-1.NumAppendedMessages': long('224634137'),
-            'kafka.logs.mytopic-1.NumberOfSegments': int('94'),
-            'kafka.logs.mytopic-1.Size': long('50143615339'),
+            'kafka.kafka.logs.mytopic-1.CurrentOffset': long('213500615'),
```
shortdudey123: Why add `kafka` again in the block of metrics? (I mean on the collector side since i realize this is the test file)

shamil: I'm not really adding `kafka`. It's taken from mbean object, you can see it on line 155 (`prefix`). This is to have a separation which metrics belong to Kafka itself and which are JVM related. Like `java.lang.GarbageCollector.*` or `kafka.consumer.*`. Those will be under a default `path`, which is also `kafka` and can be changed in the Collector configuration.

shortdudey123: > separation which metrics belong to Kafka itself and which are JVM related

Gotcha 👍

ddreggors: given the lines:
`key_prefix = prefix + "." + keys.pop(0).split('=')[-1]`
and
`keys = map(lambda i: i.replace('=', '_').replace('.', '_'), keys)`

Consider this key:
`kafka.server:type=BrokerTopicMetrics,name=BytesRejectedPerSec,topic=escher_heartbeat_event`

It sounds like we would end up with:

`kafka.kafka.server.type_BrokerTopicMetrics.name_BytesRejectedPerSec.topic_escher_heartbeat_event.Count`

is this correct?
If so then this is not preferrable, first we already start with kafka due to the path variable at the top. So we should use only the part of the prefix that we require to give needed separation (server vs kafka.server) as kafka is now redundant I think. Also I would rather remove artifacts like 'type_', 'name_', and 'topic_' as they are not needed and make your graphite paths unnecessarily long.

I supplied another patch for this here:
https://github.com/python-diamond/Diamond/issues/225

 I do like your patch as the code is more elegant than mine (non-python coder here), but would prefer to lose the parts mentioned above.

shortdudey123:

PR #456, opened 2016-04-19
Add CentOS 7 build and dev hosts to vagrant

shortdudey123:


Vagrantfile
```diff
@@ -23,23 +23,36 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
+  end
+
   config.vm.define "ubuntu-build" do |c|
     c.vm.hostname = "ubuntu-build"
     c.vm.box = "opscode_ubuntu-12.04"
     c.vm.box_url = "http://opscode-vm-bento.s3.amazonaws.com/vagrant/virtualbox/opscode_ubuntu-12.04_chef-provisionerless.box"
-  
-    c.vm.provision "shell", inline: "sudo apt-get update"  
+
+    c.vm.provision "shell", inline: "sudo apt-get update"
```
jaingaurav: Is it even a good idea to update a system's apt repo blindly? Seems like this might change more about the vagrant box than what someone may have wanted.

shortdudey123: agreed, although its not necessarily a bad thing to build stuff against what is current.
This line specifically was just removing trailing spaces, however, i can yank all system updates out if that is preferred

jaingaurav: Yeah lets do that. I don't think we should do this. Feel free to remove it either in this same commit or a second commit and I'll then pull this PR.

shortdudey123: i will add it as a 2nd commit so there is a clear separation between centos 7 and the update removal

shortdudey123: doh, i just realized this is `update` not `upgrade`
All this does is pull the latest metadata from the repos

Sure you still want it removed?

jaingaurav: Yes I still think so, applying `update` changes the package database in a manner that may be undesirable. Also, I don't think you do something similar for `yum`.

shortdudey123: Did some testing and it actually does not without without the metadata update
```
==> ubuntu-build: Err http://us.archive.ubuntu.com/ubuntu/ precise-updates/main debootstrap all 1.0.40~ubuntu0.9
==> ubuntu-build:   404  Not Found [IP: 91.189.91.23 80]
==> ubuntu-build: Get:83 http://us.archive.ubuntu.com/ubuntu/ precise/main pbuilder all 0.208ubuntu1 [356 kB]
==> ubuntu-build: Get:84 http://us.archive.ubuntu.com/ubuntu/ precise-updates/main python-magic amd64 5.09-2ubuntu0.6 [4,838 B]
==> ubuntu-build: Get:85 http://us.archive.ubuntu.com/ubuntu/ precise/main python-mock all 0.7.2-1 [15.7 kB]
==> ubuntu-build: Failed to fetch http://us.archive.ubuntu.com/ubuntu/pool/main/d/debootstrap/debootstrap_1.0.40~ubuntu0.9_all.deb  404  Not Found [IP: 91.189.91.23 80]
```

jaingaurav: Thanks for investigating. I'll merge as is.

shortdudey123: np

PR #458, opened 2016-04-21
Add memory.used_percent for RedisCollector

shortdudey123:


src/collectors/redisstat/redisstat.py
```diff
@@ -265,6 +265,24 @@ def _get_info(self, host, port, unix_socket, auth):
         del client
         return info
 
+    def _get_config(self, host, port, unix_socket, auth, config_key):
+        """Return config string from specified Redis instance and config key
+
+:param str host: redis host
```
jaingaurav: Could you please fix the indentation here?

src/collectors/redisstat/redisstat.py
```diff
@@ -265,6 +265,24 @@ def _get_info(self, host, port, unix_socket, auth):
...
+:rtype: str
+
+        """
+
+        client = self._client(host, port, unix_socket, auth)
+        if client is None:
+            return None
+
+        config_value = client.config_get(config_key)
+        del client
```
jaingaurav: Why do you need to call delete?

jaingaurav:
Looks great. If you could please address the minor comments (please ensure it is in a single squashed diff) and I'll get it merged in.

src/collectors/redisstat/redisstat.py
shortdudey123: This keeps consistency with the `_get_info` method above, but i agree w/ you

src/collectors/redisstat/redisstat.py
shortdudey123: Same as previous

shortdudey123:
@jaingaurav I can change the rest of the file to be consistent with the feedback (I just have it that way for consistency) or only my new part.  Let me know either way

jaingaurav:
That's fine, I'll pull it in as is thanks!

shortdudey123:
No problem! thanks :)

PR #460, opened 2016-04-26
Fix fluentd collector

csabiftw:


jaingaurav:
Nice catch! Any chance we could have a test case that exercises this fix?

PR #462, opened 2016-04-27
fix bad docker name join

balookrd:


jaingaurav:
Thanks, is there a reason this wasn't caught by a test case?

shortdudey123:
looks like the test case only covered what is under the container key, not the whole key namespace

jaingaurav:
Merging this in for now, but the test cases seem quite rudimentary.

PR #463, opened 2016-04-28
Add maxCollectionSize to support legacy versions of Jolokia

MichaelDoyle:
Force collection size to be unlimited for all versions of Jolokia.

From the Jolokia release notes.

> Up to 1.2.3 Jolokia truncates any collection in the response value at a threshold of 1000 elements by default. This limit can be overwritten permanently in the configuration or per request as query parameter (maxCollectionSize). However, it turned out that this limit was not large enough. So the new default behaviour is to have no limit at all. As said, if you need it you always can set a hard limit in the agent's configuration.


bmhatfield:
Thanks!

Issue #361, opened 2015-12-23
Unable to Disable DEBUG Log on Diamond


klank4135:
The only way I've found to actually disable (verbose) logging, diamond.conf

[server]
change this 
handlers = diamond.handler.graphite.GraphiteHandler, diamond.handler.archive.ArchiveHandler
to that
handlers = diamond.handler.graphite.GraphiteHandler

A possible way to logging is: log WARN+ERROR for system level diamond process things "WARNING: could not load extra python module libsomething", and ERROR for metrics "ERROR: could not connect to mysql to gather metrics".  but _not_ to log every single metric gathered... that's just way too verbose and not useful, except during debugging

adfel70:
If you are using systemd service, [this](https://github.com/python-diamond/Diamond/pull/464) should fix your problem.

shortdudey123:
Probably solved by https://github.com/python-diamond/Diamond/pull/592

PR #464, opened 2016-04-28
rpm : remove --log-stdout in systemd script

redref:
On systemd setup, with this flag enabled, log is always going to stdout (journald) whatever configuration supplied.

Both behavior are configureable (diamond.conf) without the flag.

This PR may be another way to solve #406

### DIFF
--- a/rpm/systemd/diamond.service
+++ b/rpm/systemd/diamond.service
@@ -2,7 +2,7 @@
 Description=diamond - A system statistics collector for graphite
 
 [Service]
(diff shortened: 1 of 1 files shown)


redref:
@jaingaurav 

What is your opinion on this ?

adfel70:
Maybe solve also #361

shortdudey123:
Superseded by https://github.com/python-diamond/Diamond/pull/592

PR #465, opened 2016-04-28
Fix: leading dot when prefix empty

### LINKED ISSUE #443: path_prefix and path combination producing a bug
Setting my `path_prefix` to empty and `path` to something produces an extra dot at the beginning -- at least that is what I see on the screen when I run

> diamond -f -l

which is not a huge problem since the data makes it to Graphite to the correct path, but I am not sure how exactly Graphite deals with this and whether it might be a problem in the future.

My goal was to write the Graphite metrics to a path that I completely control (without any additions by `diamond`), which -- to be honest -- was a pain and should be easier than this by default.

This is easy to fix (I think all the required changes are in `src/diamond/collector.py`, I don't code in Python and the reason I didn't fix it myself was that I found it a bit painful to test locally. I hope I'll pick this issue up myself if I can build it locally and test it, but till then, it was better to document the issue.

redref:
Quick fix #443

### DIFF
--- a/src/diamond/collector.py
+++ b/src/diamond/collector.py
@@ -351,10 +351,13 @@ class Collector(object):
         if suffix:
             prefix = '.'.join((prefix, suffix))
 
-        if path == '.':
-            return '.'.join([prefix, name])
-        else:
-            return '.'.join([prefix, path, name])
+        path_r = [name]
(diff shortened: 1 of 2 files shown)


josegonzalez:
Mind rebasing this?

PR #467, opened 2016-04-29
Add vagrant box to build docs and test collectors

shortdudey123:
This new vagrant machine (`centos7-test`) will allow for better testing of Diamond.  Different software can be installed and their respective collectors tested on the running software.


shortdudey123:
@jaingaurav can you let me know your thoughts on this one?

josegonzalez:
Thanks for the pull request!

PR #470, opened 2016-05-05
tmpfs diskspace monitored as well

anandbhoraskar:

PR #471, opened 2016-05-05
tmpfs diskspace monitored as well

anandbhoraskar:


shortdudey123:
Supersedes https://github.com/python-diamond/Diamond/pull/470

src/collectors/diskspace/diskspace.py
```diff
@@ -146,7 +146,8 @@ def get_file_systems(self):
                      mount_point.startswith('/sys'))):
                     continue
 
-                if '/' in device and mount_point.startswith('/'):
+                if ((('/' in device or device == 'tmpfs') and
```
shortdudey123: Maybe it would be better to add `tmpfs` to the `filesystems` list instead?

anandbhoraskar: the condition "'/' in device" filters it out

shortdudey123: Gotcha, `tmpfs` is still skipped though and won't make it to this line since it is not in `filesystems`

anandbhoraskar: So if the conf is modified to allow tmpfs as well, stats for /tmp will also
go through.

shortdudey123: correct

PR #473, opened 2016-05-18
Added unit awareness to client io output for CephStatsCollector

ultramathman:
The ceph -s command scales io rates, but the code was just collecting the raw digits.
Code now converts rates in the KB to EB range to bytes.

Specifically, the code in master currently sees a client io read or write rate of 1234 kB/s and stores that as 1234. It also stores 1234 if the rate is 1234 B/s or 1234 MB/s, etc. Now it stores 1234.0 if the unit is B/s, 1234000.0 if the unit is kB/s, 1234000000 if the unit is MB/s, etc.

### DIFF
--- a/src/collectors/cephstats/cephstats.py
+++ b/src/collectors/cephstats/cephstats.py
@@ -47,14 +47,14 @@ def process_ceph_status(output):
         if runit is None:
             self.log.exception('Could not get read units')
             return {}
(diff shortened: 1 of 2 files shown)


shortdudey123:
Can you fix the pep8 violations?

ultramathman:
Fixed for PEP8 and updated tests to account for my changes.

PR #474, opened 2016-05-18
Allow to use multiple sockets for getting HAProxy stats.

nihn:


src/collectors/haproxy/haproxy.py
```diff
@@ -110,12 +110,13 @@ def http_get_csv_data(self, section=None):
                            "(Invalid username or password?) %s", e)
             return metrics
 
-    def unix_get_csv_data(self):
+    def unix_get_csv_data(self, section=None):
+
```
shortdudey123: Extra newline

PR #477, opened 2016-05-27
Add Basic/Shield auth to elasticsearch collector

okushchenko:
Current implementation of Elasticsearch collector has no ability to authenticate against Shield plugin for Elasticsearch, so I've added optional parameters to authenticate with.
I'm using this version of collector in prod right now and it works just fine.

### DIFF
--- a/docs/collectors/ElasticSearchCollector.md
+++ b/docs/collectors/ElasticSearchCollector.md
@@ -27,12 +27,14 @@ logstash_mode | False | If 'indices' stats are gathered, remove the YYYY.MM.DD s
 measure_collector_time | False | Collect the collector run time in ms | bool
 metrics_blacklist | None | Regex to match metrics to block. Mutually exclusive with metrics_whitelist | NoneType
 metrics_whitelist | None | Regex to match metrics to transmit. Mutually exclusive with metrics_blacklist | NoneType
+password |  | Password for Basic/Shield auth | str
 port | 9200 |  | int
 stats | jvm, thread_pool, indices, | Available stats:<br>
  - jvm (JVM information)<br>
  - thread_pool (Thread pool information)<br>
  - indices (Individual index stats)<br>
  | list
+user |  | Username for Basic/Shield auth | str
 
 #### Example Output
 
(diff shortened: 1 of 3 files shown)


shortdudey123:
Can you update the testing and doc as well?

okushchenko:
@shortdudey123 Can you help me with the testing part - what tests should I write?

shortdudey123:
I would duplicate one of the `real_data` tests and add in the username / password auth in the mock and make sure that it passes those variables in the request and gets the data back

hamelg:
I need this enhancement. What about this test : 
```
--- a/src/collectors/elasticsearch/test/testelasticsearch.py
+++ b/src/collectors/elasticsearch/test/testelasticsearch.py
@@ -119,6 +119,12 @@ class TestElasticSearchCollector(CollectorTestCase):
         self.assertPublishedMany(publish_mock, metrics)
 
     @patch.object(Collector, 'publish')
+    def test_should_work_with_real_data_and_basic_auth(self, publish_mock):
+        self.collector.config["user"] = "user"
+        self.collector.config["password"] = "password"
+        self.test_should_work_with_real_data()
+
+    @patch.object(Collector, 'publish')
     def test_should_work_with_real_data_logstash_mode(self, publish_mock):
         returns = [
             self.getFixture('stats'),
```

shortdudey123:
@hamelg looks like that would probably work if not close to it

hamelg:
What's missing to accept this request ?

shortdudey123:
I still don't see any testing or doc as requested

hamelg:
For testing, see my comment https://github.com/python-diamond/Diamond/pull/477#issuecomment-267843330
For doc, adding user/password in config_help.update() is not sufficient ?

I can't amend this request and okushchenko has given up.
How can I do ?
May i recreate another request ?

shortdudey123:
> For testing, see my comment #477 (comment)

This should be added to the PR, not a comment

> For doc, adding user/password in config_help.update() is not sufficient ?

`docs/collectors/ElasticSearchCollector.md` needs to be updated

@okushchenko can you update as requested?

kormoc:
Actually, `make docs` should update the fines in the `docs/` directory

https://github.com/python-diamond/Diamond/blob/master/Makefile#L39

okushchenko:
I've incorporated a test proposed by @hamelg, many thanks! Also, I've updated the docs.

okushchenko:
@shortdudey123 is there anything more I can do to make this PR mergeable?

PR #478, opened 2016-05-27
Force xenfix as opt-in

tiras-j:
The bug in Xen which this fix was accommodating was fixed in 2012.
Currently all AWS machines use Xen hypervisors which are already fixed,
however because of the aggressive behavior of the "xenfix" in this collector, any machines that
are running Diamond on AWS hosts which have not actively configured this
setting are reporting incorrect CPU idle values. This is unfortunate as
this value is often used for provisioning cluster sizes and the like.

Optionally it's likely safe to remove the code altogether, as anyone running an old Xen hypervisor likely isn't keeping their Diamond core up-to-date. But who knows the unique use-cases of certain situations.

EDIT: 
The original bug report: https://bugzilla.redhat.com/show_bug.cgi?id=624756
The fix (2012): http://rhn.redhat.com/errata/RHSA-2012-0862.html

### DIFF
--- a/src/collectors/cpu/cpu.py
+++ b/src/collectors/cpu/cpu.py
@@ -55,7 +55,7 @@ class CPUCollector(diamond.collector.Collector):
         config.update({
             'path':     'cpu',
             'percore':  'True',
-            'xenfix':   None,
+            'xenfix':   'False',
             'simple':   'False',
             'normalize': 'False',
         })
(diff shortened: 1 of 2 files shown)


shortdudey123:
> any machines that are running Diamond on AWS hosts which have not actively configured this setting are reporting incorrect CPU idle values

Can you elaborate on this?  I am not seeing this on my hosts and i don't have `xenfix` set.  Also, if the test is working correctly, then you should be getting the correct values

tiras-j:
@shortdudey123 Check the aggregate of the cpu.* values emitted via the CPU collector per interval, do they add up to the expected value based on the systems configured USER_HZ?

PR #482, opened 2016-06-07
Exclude devices from disk usage statistics

owahlberg:
Added a simple option to exclude devices or partitions based on a regular expression.
I use this to exclude very specific devices without complicating the default devices regex.


shortdudey123:
Same as https://github.com/python-diamond/Diamond/pull/444

shortdudey123:
Can you update the tests and doc?

shortdudey123:
This is a subset of the commits in https://github.com/python-diamond/Diamond/pull/483

PR #483, opened 2016-06-07
Add support for persistent disk naming for DiskUsage collector

owahlberg:
This patch adds support for using ID_MODEL and ID_SERIAL for persistent naming of devices.
Very useful in large disk environments to get consistent naming.


shortdudey123:
Can you remove the overlap with https://github.com/python-diamond/Diamond/pull/482 ?

PR #484, opened 2016-06-07
Add collector for Linux kernel bcache statistics.

owahlberg:
Rudimentary bcache collector.
Collects cache hit ratio, hit/miss/bypass/readahead statistics.
Based on https://github.com/pommi/collectd-bcache/


shortdudey123:
Can you fix the pep8 errors, add testing, and docs?

PR #487, opened 2016-06-09
Collect diskusage statistics for FusionIO Cards.

reyjrar:
We're using FusionIO Cards for things and the diskusage collector doesn't collect that data due to the regex in use.  This patch adds that functionality.


shortdudey123:
Can you update the doc? `docs/collectors/DiskUsageCollector.md `

shortdudey123:
The help section should also mention `fio`
https://github.com/python-diamond/Diamond/pull/487/files#diff-b9243715c2d8c420469e67f2dd23cce1R47

faja:
Bear in mind you need to update sector size to 4K to collect fusion IO usage stats.
If you have standard (sector size=512b) disk and fio you probably need two separate collectors.

PR #491, opened 2016-06-16
Fix docker collector exception handling

csabiftw:
SIGALRMException was handled on the collector level, resulting in a long stacktrace printed when the collector timed out. Scheduler already takes care of exceptions coming from collector, so removing the try-except block from DockerCollector.


jpiwonski:
+1

PR #494, opened 2016-06-17
graphite handler: add parameter to control interval reconnecting

rtoma:
Default behaviour (neverending connections) is unchanged.

Setting this parameter allows you to re-balance the load on a loadbalanced cluster after node maintenance or an incident.  Without this feature you would have all Diamond instances unevenly connected to the Graphite nodes, until you restart all Diamond instances.

### DIFF
--- a/docs/handlers/GraphiteHandler.md
+++ b/docs/handlers/GraphiteHandler.md
@@ -34,3 +34,4 @@ scope_id | 0 | IPv6 Scope ID | int
 server_error_interval | 120 | How frequently to send repeated server errors | int
 timeout | 15 |  | int
 trim_backlog_multiplier | 4 | Trim down how many batches | int
+reconnect_interval | 0 | How often (in seconds) to reconnect to graphite. Default (0) is never | int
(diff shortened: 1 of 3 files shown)


shortdudey123:
Looking good :)
I would squash your commits before merge though

rtoma:
@shortdudey123 squashed the commits. feel free to merge

shortdudey123:
thanks! forgot to mention that the doc should be updated as well

rtoma:
thanks for reminding me, @shortdudey123: docs for the graphitehandler have been added

jaingaurav:
Thanks @rtoma looks great!

scottcunningham:
It looks like this works for the GraphitePickleHandler too (since GraphitePickleHandler is a subclass of GraphiteHandler). Would it make sense to document this new config option for the Pickle handler also, or is it preferred to just refer to the GraphiteHandler docs? I could see this working two ways:

- Explicitly document new config options for GraphiteHandler in both GraphiteHandler and GraphitePickleHandler docs
- Add a note to the PickleHandler docs referring to the GraphiteHandler docs with a note that GraphiteHandler config options are also valid

What are your thoughts?

PR #490, opened 2016-06-15
Adding trove classifiers.

nickjacobson:

PR #495, opened 2016-06-17
Adding trove classifiers.

nickjacobson:
Adding trove classifiers for Python 2 only.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -144,5 +144,9 @@ setup(
     scripts=['bin/diamond', 'bin/diamond-setup'],
     data_files=data_files,
     install_requires=install_requires,
+    classifiers=[
+        'Programming Language :: Python',
+        'Programming Language :: Python :: 2',
+        ],
     ** setup_kwargs
 )
--- a/setup.py
+++ b/setup.py
@@ -145,8 +145,8 @@ setup(
     data_files=data_files,
     install_requires=install_requires,
     classifiers=[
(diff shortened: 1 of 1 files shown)


shortdudey123:
Replaces https://github.com/python-diamond/Diamond/pull/490

setup.py
```diff
@@ -144,5 +144,9 @@ def pkgPath(root, path, rpath="/"):
     scripts=['bin/diamond', 'bin/diamond-setup'],
     data_files=data_files,
     install_requires=install_requires,
+    classifiers=[
+        'Programming Language :: Python',
+        'Programming Language :: Python :: 2',
+        ],
```
shortdudey123: Indent this to the same level as `classifiers` for consistency within the file

nickjacobson: I tried it and that broke the regression tests, so I reverted the indentation level back.

shortdudey123: hmm interesting
it doesn't complain [here](https://github.com/nickjacobson/Diamond/blob/c1e74905303bc3f4d6d2ce191212b2c923e14694/setup.py#L36) or [here](https://github.com/nickjacobson/Diamond/blob/c1e74905303bc3f4d6d2ce191212b2c923e14694/setup.py#L43)

nickjacobson: ok, you meant to just dedent the line with the bracket, doing so.

josegonzalez:
Thanks for the pull request!

Issue #499, opened 2016-07-01
Diskspace Collector incorrectly calculating bytes on OSX Yosemite


Ssawa:
Hey,

I've been evaluating Diamond on my local OSX machine and noticed that the DiskSpace collector was giving me bogus results (saying I had ~3000 GB free on my 250 GB hard drive). Strangely, however, I noticed that the "percentfree" metric looked pretty much spot on. After digging into the code it looks like the problem was where we [set block_size to f_bsize](https://github.com/python-diamond/Diamond/blob/master/src/collectors/diskspace/diskspace.py#L215), after consulting a [few other resources](http://stackoverflow.com/questions/4260116/find-size-and-free-space-of-the-filesystem-containing-a-given-file) I replaced block_size with f_frsize and was getting the proper calculations.

It looks like on some systems (tested on Ubuntu at least) f_bsize and f_frsize are the same but on my Yosemite machine f_bsize was "1048576" while f_frsize was the much more reasonable "4096". I decided to open up a ticket rather than submit a pull request because to be perfectly honest I don't know too much about the differences between f_bsize and f_frsize and don't know what all ramifications of the change might be, however I will be making the change on my own fork.

Ssawa:
Also if[ this forum post](http://community.hpe.com/t5/LVM-and-VxVM/definition-of-f-bsize-vs-f-frsize/td-p/4761159) is any indication, it seems like the f_bsize vs. f_frsize confusion has come up before.

josegonzalez:
Thanks for the pull request!

PR #503, opened 2016-07-07
Diskusage Collector: Change block_size from f_bsize to f_frsize

Ssawa:
Fix for issue #499

### DIFF
--- a/src/collectors/diskspace/diskspace.py
+++ b/src/collectors/diskspace/diskspace.py
@@ -212,7 +212,10 @@ class DiskSpaceCollector(diamond.collector.Collector):
                     self.log.exception(e)
                     continue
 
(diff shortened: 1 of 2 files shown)


jaingaurav:
Thanks @Ssawa this looks really good. Could you 2 additional things:
1. Add a comment above the line you changed explaining the reason for the change
2. Add a test case. Look at `testdiskspace.py` and `test_should_work_with_real_data`. You could simply make `f_bsize = 1048576` in that test case and it'd be fine since that would demonstrate that `f_bisze` and `f_frsize` can be different.

Please be sure to squash all the commits and upload once you are done and I'll pull this in.

jaingaurav:
Thanks!

PR #505, opened 2016-07-20
Add name/value pair generic file collector.

stantona:
The intention of this collector is to gather metrics found
in files. Candidates could be standard proc files such as meminfo or 
vmstat but could be used for any file that follows a name/value format.

Given a regex and a file with metrics, the collector can extract
the name of a metric and its value. It assumes that each line follows
the same format (see vmstat or meminfo). It is very similar to how many
of use awk to extract such information.

This was written in response to the current vmstat collector
limiting the metrics to a very small subset. The idea is that this
can be used for any file, or can be extended to customize behaviour.
This could form the base class for many existing collectors. 

The two required config options are 'regex' (which has a default value),
and 'file' (with no default value).


src/collectors/genericfile/genericfile.py
```diff
@@ -0,0 +1,84 @@
...
+        return config_help
+
+    def get_default_config(self):
+        """
+        Returns the default collector settings
+        """
+        config = super(GenericFileCollector, self).get_default_config()
+        config.update({
+            'regex': '^\s*(?P<name>\w+)\s*[\s|:|=]\s*(?P<value>\w+)$',
+            'path': '',
```
shortdudey123: Not sure the top level is the best place for this

stantona: Yeah. I'm not sure what a good default path would be as the path will be dependent on the file. For example, if you're working with '/proc/meminfo', you could set the path to 'meminfo'.  I could give it some generic name, or remove it and treat it as a mandatory option that the user must set in the configuration.

shortdudey123: maybe `generic_file` for a default?

src/collectors/genericfile/genericfile.py
```diff
@@ -0,0 +1,85 @@
...
+        return config_help
+
+    def get_default_config(self):
+        """
+        Returns the default collector settings
+        """
+        config = super(GenericFileCollector, self).get_default_config()
+        config.update({
+            'regex': '^\s*(?P<key>\w+)\s*[\s|:|=]\s*(?P<value>\w+)\s*$',
+            'path': '',
```
shortdudey123: Any reason to yank the default of `generic_file` back out?

stantona: Ahh, my bad. I'll fix that.

shortdudey123: any update on this?

Issue #66, opened 2015-01-16
network collector: vlan interfaces with dots


agaoglu:
Some of our servers use 8021q vlan interfaces and are named like eth2.23

These create metrics such as

```
servers.server1.network.eth2.23.rx_byte
servers.server1.network.eth2.23.tx_byte
servers.server1.network.eth2.rx_byte
servers.server1.network.eth2.tx_byte
```

causing metric path structure to change. It'd be better if the dot was replaced with something like _.

josegonzalez:
There was an issue raised in the old repo regarding this, let me see if I can find it...

shortdudey123:
Waiting for tests to be added in https://github.com/python-diamond/Diamond/pull/507

PR #507, opened 2016-07-27
network collector: sanitize device names with dots

jnoss:
This sanitizes network device names (such as vlans) containing '.' to
have '_' instead, to preserve metric path structure. Addresses issue #66


shortdudey123:
Can you update the tests to make sure this is tested?

PR #509, opened 2016-07-29
Allow precision to be set in nginx collector

scottcunningham:
This patch allows us to set precision in the Nginx collector's config file. Eg:

```
enabled = True
precision = 2
req_port = 9080
```

Changing the precision from 0 to 2 resulted in this change on a testing system when viewing the data in Grafana:

asdf

Note that the number of requests/sec hasn't changed, just the precision config value.

When you hover over a datapoint in Grafana:
- Some point in time before the patch:
  
  asdf
- After the patch and changing the config value (with requests coming in at the same rate):
  
  asdf

### DIFF
--- a/src/collectors/nginx/nginx.py
+++ b/src/collectors/nginx/nginx.py
@@ -37,6 +37,7 @@ class NginxCollector(diamond.collector.Collector):
     def get_default_config_help(self):
         config_help = super(NginxCollector, self).get_default_config_help()
         config_help.update({
+            'precision': 'Number of decimal places to report to',
             'req_host': 'Hostname',
             'req_port': 'Port',
             'req_path': 'Path',
@@ -47,6 +48,7 @@ class NginxCollector(diamond.collector.Collector):
 
     def get_default_config(self):
         default_config = super(NginxCollector, self).get_default_config()
+        default_config['precision'] = 0
         default_config['req_host'] = 'localhost'
         default_config['req_port'] = 8080
         default_config['req_path'] = '/nginx_status'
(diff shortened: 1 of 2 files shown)


scottcunningham:
Oh dear, failing tests! They look like some pretty obvious PEP-8 violations. I'll get them fixed up right now.

As an aside, are there any docs I can refer to about running the tests locally?

src/collectors/nginx/nginx.py
```diff
@@ -47,6 +47,7 @@ def get_default_config_help(self):
 
     def get_default_config(self):
         default_config = super(NginxCollector, self).get_default_config()
+        default_config['precision'] = 0
```
shortdudey123: add to the `config_help` section to

shortdudey123:
Doc needs to be updated too

scottcunningham:
Thanks for the feedback! I'm away from a computer for the weekend but will address the issues on Monday.

On 29 Jul 2016 18:55, Grant Ridder <notifications@github.com> wrote:

Doc needs to be updated too

scottcunningham:
Docs and config_help updated. I used the same wording for precision as in the NTP collector.

shortdudey123:
LGTM
@jaingaurav can you review?

scottcunningham:
Hey @shortdudey123, @jaingaurav, do you know when you'll have a chance to take another look at this? Thanks!

shortdudey123:
Already looks good to me
Waiting on @jaingaurav or another python-diamond member to merge

scottcunningham:
@jaingaurav Any chance you could take a quick look at this please? It's been ready and waiting for a few months now.

shortdudey123:
@bmhatfield can you review when you have a change? :)

rowleyaj:
@bmhatfield @shortdudey123 @josegonzalez Any updates on what's needed to get this PR merged. This is now the only custom part of the diamond install that we have.

josegonzalez:
Seems legit, will go out in the next release when I have a moment.

PR #513, opened 2016-08-08
Valuable stats are missed due to missing types

ddreggors:
Some stats types are labeld as 'java.lang.Object' in MX4J.
These stats are valuabe and get missed because we check for 'type' and
only use it if it matches double, float, long, or int.
Because of this we seem to miss all FetcherLagMetrics.


shortdudey123:

Issue #519, opened 2016-08-15
SNMPRawCollector fails due to unexpected response


dwagon:
Using the SNMPRawCollector fails as the data returned by the SNMPCollector() class pretties up the data in the format: SNMPv2-SMI::enterprises.1.2.3.... instead of 1.3.6.1.4.1.1.2.3...

This causes the collector to complain about unexpected response because it isn't in the format that was requested.

jacksu:
Which file call the function collect_snmp(self, device, host, port, community) in the file snmpinterface.py?

dwagon:
snmpinterface collector isn't involved. This is the snmpraw collector.

josegonzalez:
Closed by #520.

PR #520, opened 2016-08-15
Return numeric OID for SNMP - not pretty version

dwagon:
Using the SNMPRawCollector fails because it is expecting the return from the snmp collector in numeric OID not prettied up OID:

This solves #519


josegonzalez:
Can you provide tests so that this doesn't recur?

josegonzalez:
Merged, thanks for the pull request (and sorry for the wait!).

PR #521, opened 2016-08-15
tsdbhandler tag handling

kullervo16:
I had some issues using the tags clause of the tsdbhandler.. if you do not know the format in which to specify the tags, the handler will simply not send any data (and not log any error either).... so you had to look into the OpenTSDB documentation to find out that tags are key value pairs, separated by spaces... and since the tags are appended to the default tag, it should start with a space...

Changed this to:
- overcome the missing space 
- allow both comma separated and space separated tags
- documented the format, the default tag, added an example and a link to the proper openTSDB page.

I also added some tests to test this behaviour (and hence also the basic tsdb collector behaviour)


shortdudey123:
Can you fix all the pep8 issues?

shortdudey123:
Subset of https://github.com/python-diamond/Diamond/pull/528

PR #527, opened 2016-09-01
add https support to elasticsearch collector without breaking existing configs

imreACTmd:
ElasticSearch now has SSL/TLS/HTTPS support via several means: Shield, SearchGuard. 

This patch allows specifying the connection scheme which was previously hardcoded as HTTP. 

The default scheme is still HTTP but now it can also be specified as HTTPS.


shortdudey123:
Can you fix the pep8 error?
`src/collectors/elasticsearch/elasticsearch.py:229:81: E501 line too long (87 > 80 characters)`


Also, can you update the doc with the new config option?

imreFitos:
Thank you for the quick feedback, updated PR

shortdudey123:
LGTM

@jaingaurav can you review?

bmhatfield:
Thanks for this change!

PR #528, opened 2016-09-05
TSDB metric tags

kullervo16:
After working with the tags on the version I proposed in PR #521, I ran into some trouble with an Openshift HAProxy instance that I was handling... since this proxy contains lots of single server backends, I got thousands of metrics that were highly redundant (the server counter + the aggregate). Furthermore, since the backend servers names are dynamical, I get 1000s of metrics that live for some minutes, and never will be used again... This really screws up code completion when you want to work with the data.

Reading [Metrics vs Tags](http://opentsdb.net/docs/build/html/user_guide/query/timeseries.html) in the OpenTSDB documentation rang a bell : the structural data in the metric name should partly be converted to tags. As it also states, you should not force aggregations into a metric value (like total) since the backend is capable of calculating that for you (which removes the need of about 50% of the counters in my case)

So I added support for this. There problem is that to achieve this, either the collector needs to know the capabilities of the handler, or the handler needs to understand the structure of the collector... both are a bit of a nuissance, but I think the latter was the cleanest.

The current implementation "knows" the structure of 3 collectors (cpu, HAProxy and a new one that I will be committing later : mattermost). For these the structural parts are converted to tags (and removed from the metric path). It also marks the aggregates, which you can skip based on configuration.

The result is that I now have only 30 counters, which is much cleaner in the grafana code completion, and the templating mechanism for opentsdb works as it should.

Metrics from other collectors remain unchanged (but can be added with minimal effort). If you still want the old way, you can simply deactivate the tag calculation and/or aggregate skipping. Since I think the new setting is more userfriendly, I used it as default.

You can have a look at the tests to see the cleaning in action. Comments are welcome.


shortdudey123:
@kullervo16 can you squash your commits a bit to remove all of the pep8 fix commits?

PR #531, opened 2016-09-09
Debian package. Remove deprecated python-support package

InHavk:
The python-support is deprecated and not support in Ubuntu 16.04+ in favor of dh_python2


debian/control
```diff
@@ -5,8 +5,9 @@ Maintainer: Rob Smith <kormoc@gmail.com>
 Homepage: https://github.com/BrightcoveOS/Diamond
 Vcs-Git: git://github.com/BrightcoveOS/Diamond.git
 Vcs-Browser: https://github.com/BrightcoveOS/Diamond
-Build-Depends: debhelper (>= 7), python (>= 2.4), python-support, python-mock, python-configobj, cdbs
+Build-Depends: debhelper (>= 7), python-all, python-setuptools, dh-python, python-mock, python-configobj, cdbs
```
josegonzalez: Can we instead do:

```
python-support | dh-python
```

That would allow us to support older versions I think (assuming dh-python isn't there).

debian/rules
```diff
@@ -1,11 +1,13 @@
 #!/usr/bin/make -f
 # -*- makefile -*-
 
-DEB_PYTHON_SYSTEM := pysupport
-
-include /usr/share/cdbs/1/rules/debhelper.mk
-include /usr/share/cdbs/1/class/python-distutils.mk
+#include /usr/share/cdbs/1/rules/debhelper.mk
+#include /usr/share/cdbs/1/class/python-distutils.mk
```
josegonzalez: Why did you comment it out?

josegonzalez:
*bump* :)

debian/rules
cbaenziger: With out this I see one ends up with the following failure:
```
dh debian/_cdbs_buildflags.mk --with python3 --buildsystem=pybuild
dh: Unknown sequence debian/_cdbs_buildflags.mk (choose from: binary binary-arch binary-indep build build-arch build-indep clean install install-arch install-indep)
test -x debian/rules
dh reverse-config --with python3 --buildsystem=pybuild
dh: Unknown sequence reverse-config (choose from: binary binary-arch binary-indep build build-arch build-indep clean install install-arch install-indep)
```

I have not done much at all with Debian packaging so I do not quite appreciate this significance but I am failing to find a way to build this using both `python-support` and that "era" of rules file and `dh_python` and that style? (However I am similarly broken needing to build on Ubuntu 16.04+.)

Is it required that Diamond still build on such old versions of Ubuntu or can 16.04 become the new lowest version of support?

shortdudey123:
Taken care of by #710

PR #532, opened 2016-09-13
I have tested. Snmp protocol supports now.

jacksu:


shortdudey123:
Mostly a dupe of https://github.com/python-diamond/Diamond/pull/310

josegonzalez:
This was resolved via the other two pull requests. Thanks for keeping on top of it!

PR #534, opened 2016-09-21
make userscripts collector error messages sane

keen99:
we have 2 almost identical error outputs - one when stderr contains something, and another when stdout contains something we couldn't split.   if you don't notice the "d" difference, debugging the result is near impossible.

changed the errors to reflect the actual conditions.


shortdudey123:
👍

josegonzalez:
Merged, thanks for the pull request!

PR #536, opened 2016-09-22
fix error handling output for userscripts collector

keen99:
another fix for error cases:  

before: if stderr is returned, but no stdout, we bail out before we report the stderr content - throwing away the error message that may only be reproducible in the conditions of the collector.

after:  if stderr is returned, log it regardless of if we bail.   log it as warning, not error, since it's non-blocking.
 
also removed "skipping" from the proc.returncode - we dont actually bail unless there's no stdout, regardless of return code, so dont tell the user we are.


josegonzalez:
Seems like there are some pep8 test fails here, mind fixing those up?

PR #544, opened 2016-10-05
calculate available space based on available blocks instead of free blocks

SvenSchindler:
Hi, the current computation for available space uses the amount of free blocks instead of available blocks which may not always be a good idea because free blocks include blocks which can exclusively be used by root.


kormoc:
Thank's so much for this! I'm just wondering if we should leave this metric path alone and add a new one for the available amount of free blocks?

shortdudey123:
I would tend to agree w/ @kormoc since that keeps this from being a breaking change

PR #545, opened 2016-10-21
fix puppetdb collector, so it can work on all version of puppetdb

ashish1099:
Old puppetdb.py only handles puppetdb 2x. 
This patch will support puppetdb 3.x and puppetdb 4.x


josegonzalez:
This needs tests for all versions of puppetdb before it can be merged.

PR #548, opened 2016-11-18
fix #547 MongoDB collector reports partial metrics

### LINKED ISSUE #547: MongoDB collector reports partial metrics
MongoDB metrics are taken from `db.serverStatus` and lots of metric names contain white spaces.
Therefore they are not published on Graphite.
For instance: `wiredTiger.cache.eviction server evicting pages` and so on.
A PR is following.
Regards
ben

bdelbosc:
Sanitizing the metrics names.


shortdudey123:
Can you update the tests to cover this case so that a regression does not happen

josegonzalez:
Closes #547

jaingaurav:
Looks great merged!

PR #550, opened 2016-11-25
docs: Add release process notes

josegonzalez:
@jaingaurav Thoughts? I figure adding them here seems reasonable (I'd also like to get us on that regular release cadence).


docs/Development/Release-Process.md
```diff
@@ -0,0 +1,62 @@
...
+
+## pypi
+
+The pypi package can be found at https://pypi.python.org/pypi/diamond
+
+Currently the three maintainers are:
+
+- kormoc
+- mattrobenolt
+- savant
```
jaingaurav: Could you include me here. I believe I have acce to upload releases

docs/Development/Release-Process.md
```diff
@@ -0,0 +1,62 @@
...
+The major and minor version numbers are generated based on the latest annotated tag.
+
+Cutting a new major/minor version is done via:
+
+```shell
+git tag -a v4.1 -m "foobar"
+git push origin --tags
+```
+
+From this point forward, the makefile will generate a version starting with 4.1
```
jaingaurav: I think we need some guidelines for versioning. How about x.x.x:
Major - Major features, possible backwards incompatible changes, etc.
Minor - New collectors/handlers
Patch - Minor bug fixes, collector tweaks

What do you think? Would the minor number bump up too often with this scheme?

docs/Development/Release-Process.md
josegonzalez: Will do.

docs/Development/Release-Process.md
josegonzalez: That seems reasonable to me. Will modify the pull request.

josegonzalez:
One thing I'm not sure on how to do this, but how should we generate docs locally? Running `make docs` breaks without a `conf/diamond.conf` in the path, and if it's not installed. That seems... wrong?

jaingaurav:
So the docs are managed by readthedocs right?

But yeah I agree that running make should just work. I think we shouldn't rely on any config file for the building of anything.

josegonzalez:
Right but we generate md files for each collector, which I think is done via `make docs`.

@shortdudey123 Any thoughts here? I've noticed you commenting on issues in the past :)

shortdudey123:
When i generate docs, i do it inside the vagrant box that got merged in https://github.com/python-diamond/Diamond/pull/467

shortdudey123:
With this PR in general though, it looks like it describes the problem of commits since the last tag, but does not appear to solve that problem

josegonzalez:
@shortdudey123 what would you change? My goal is to get us onto a two-week release cadence, with packaging available for all stable LTS distros.

shortdudey123:
If the intention is to switch to semver, then i would say that needs to be explicitly called out.  Looks like @jaingaurav added a comment about it already :)

docs/Development/Release-Process.md
kormoc: Sorry, I wasn't clear. If we split up all collectors/handlers into their own repos, what is a minor point release of the core at that point? Or do we bring them all together to package them so it's purely just a code artifact that they're split up?

josegonzalez: I'm not sure how a splitup works, but in theory if we add new - non-breaking - functionality to the core, thats a minor release?

shortdudey123:
@jaingaurav @josegonzalez further thoughts?

PR #552, opened 2016-11-25
Set a proper email address in setup.py

josegonzalez:
This should fix issues uploading the package to twine, as per https://github.com/h5py/h5py/commit/0efaceeffb3693a4dd1ef0cd60b93d74c870cee0

/cc @kormoc @jaingaurav in case you have an email you'd prefer this go this. For me, it's just a catchall.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -136,7 +136,7 @@ setup(
     version=version,
     url='https://github.com/python-diamond/Diamond',
     author='The Diamond Team',
-    author_email='https://github.com/python-diamond/Diamond',
+    author_email='diamond@librelist.com',
     license='MIT License',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -136,7 +136,7 @@ def pkgPath(root, path, rpath="/"):
     version=version,
     url='https://github.com/python-diamond/Diamond',
     author='The Diamond Team',
-    author_email='https://github.com/python-diamond/Diamond',
+    author_email='pypi@josediazgonzalez.com',
```
jaingaurav: I'm happy to add an e-mail but it'd be nice if GitHub or pypi supported some sort of proxy e-mail address or perhaps point to a mailing list instead?

PR #555, opened 2016-11-29
Change deprecated psutil memory_info_ex to memory_info

aivarasko:
psutil memory_info_ex method is deprecated:

https://pythonhosted.org/psutil/#psutil.Process.memory_info_ex

This PR changes memory_info_ex to memory_info.


kormoc:
So previous to 4.0.0, memory_info doesn't contain all the same data as memory_info_ex. Can we handle this in a backwards comparable way and use memory_info_ex if it exists or fall back to memory_info?

PR #556, opened 2016-12-08
Add memcached slab collector.

dellis23:


src/collectors/memcached_slab/test/testmemcached_slab.py
```diff
@@ -0,0 +1,45 @@
...
+import sys
+import unittest
+
+# Make sure we can import the collector, which lives in the directory above.
+sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+import memcached_slab
+
+
+RAW_SLAB_STATS = """STAT 1:chunk_size 96\r
```
shortdudey123: Can you put this in a fixture file instead?

dellis23:
@shortdudey123 This generates a lot of additional stats folks might not want to have loaded in their stats cluster.  It's also an entirely different call to get the stats back.  That's probably not as big of a deal, but it makes me lean towards these being two separate levels of introspection that should probably only be added if the user explicitly requests them.  I'd hate for a bunch of people to update diamond and find that their graphite nodes unexpectedly out of disk.

Thanks for the heads up re: fixtures and docs, I'll do that now.

shortdudey123:
Fair point

Maybe make it a boolean config option on the `MemcachedCollector`? a bit of the code appears to be reused between the too.  Just a thought though, if it doesn't make sense to do that, ignore that :p

dellis23:
@shortdudey123 I've updated the tests with fixtures, but I can't find any docs on how to generated the markdown docs.  Can you point me in the right direction?

dellis23:
Apologies, I see some stuff for it in the makefile.  I'll try that.

dellis23:
@shortdudey123 Alright, should be good to go.  Looks like the build is failing for all PRs right now -- unable to find cython.

Issue #218, opened 2015-07-08
Mismatch between docs and behavior of PingCollector


aboks:
Suppose `PingCollector` is configured with

```
target_website = 'www.example.org'
```

The documentation ("We extract out the key after target_ and use it in the graphite node we push.") suggests that the metric sent will be `servers.hostname.ping.website`, but the actual metric sent is `servers.hostname.ping.www_example_org`. I don't know what the intended behavior is (and thus whether the doc or the code is wrong), only that there is a mismatch.

josegonzalez:
It uses the target as the metric name. It wouldn't chomp off the subdomain - `www` - or the tld - `org` - as that would have the possibility of causing overlapping metric collection.

As an example, the following would collide:

```
admin.example.com
example.org
```

Those are obviously two different domains, but by your logic, we'd have them both report to `example`, which is definitely wrong. Instead, we simply replace periods in the key to create the metric name.

Hope that helps.

aboks:
Thank you for looking into this @josegonzalez, but I think you misunderstood the issue. I did not mean to suggest that the implementation should strip the subdomain and TLD, as that would obviously be a bad idea for the reasons you pointed out.

Rather, I see there is a mismatch between the documentation (which suggests that the part of the setting name after `target_` is pushed) and the actual code (which sends the hostname with dots replaced by underscores). So either:
* The desired behavior in my example is to push `www_example_com`. In that case the documentation is incorrect and should be updated to say that the metric is derived entirely from the hostname and the part of the setting name after `target_` has no influence...
* ... or the desired behavior in my example is to push `website` (*not* `example`!), as the documentation says. In that case the implementation is incorrect and should probably be fixed.

josegonzalez:
Ah good catch. Mind digging through the ping collector history to figure out what exactly we originally meant to do here?

aboks:
The current behavior (replacing dots with underscores in the hostname) seems to be there from the initial commit of the `PingCollector` (https://github.com/python-diamond/Diamond/commit/be5b89d753ea90b9fe07d70d768cd72cc2c92492). The remark in the documentation was added in https://github.com/python-diamond/Diamond/commit/d62087af7e05412ce7d1cc0ac7a36d2e5c4fcebe, but did not match the implementation at that time. So it was either misunderstanding of the implementation, wishful thinking, or some change that was planned but forgotten.

[...]

PR #559, opened 2016-12-14
[Ping] Fix docs for actual behavior

aboks:
Fixed incorrect description of how the graphite nodes are derived from the 
configuration (see https://github.com/python-diamond/Diamond/issues/218). Documented actual behavior instead.

### DIFF
--- a/docs/collectors/PingCollector.md
+++ b/docs/collectors/PingCollector.md
@@ -32,7 +32,8 @@ for your targets in pairs like:
 
 'target_1': 'example.org'
 
-We extract out the key after target_ and use it in the graphite node we push.
+The graphite nodes pushed are derived from the pinged hostnames by replacing all
+dots with underscores, i.e. 'www.example.org' becomes 'www_example_org'.
 
 
 #### Options
@@ -52,5 +53,7 @@ use_sudo | False | Use sudo? | bool
 
 ```
 servers.hostname.ping.localhost 11
+servers.hostname.ping.www_example_org 23
+servers.hostname.ping.192_168_0_1 16
 ```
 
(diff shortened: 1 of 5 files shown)


shortdudey123:
Can you update the actual doc file as well?

aboks:
Ah, I didn't know there was one, thought the docs were generated directly from the docstrings. Also updated the doc file now.

docs/collectors/PingCollector.md
```diff
@@ -52,5 +53,7 @@ use_sudo | False | Use sudo? | bool
 
 ```
 servers.hostname.ping.localhost 11
+servers.hostname.ping.www_example_org 23
```
shortdudey123: This section of the doc file is generated from the `test_should_work_with_real_data_host_gentoo` method in the test file so this section is actually correct.  If you want to add more examples, please update the test method.

src/collectors/ping/test/testping.py
```diff
@@ -48,18 +48,44 @@ def test_should_work_with_real_data_bad_gentoo(self, publish_mock):
...
+            'target_c': '192.168.0.1',
+            'bin': 'true'
+        })
 
-        patch_communicate.start()
-        self.collector.collect()
-        patch_communicate.stop()
+        collector = PingCollector(config, None)
+
+        def popen_se(command, **kwargs):
```
shortdudey123: Not sure what this is needed for
See https://github.com/python-diamond/Diamond/blob/master/src/collectors/elasticsearch/test/testelasticsearch.py#L55-L61

aboks: It is needed because the order in which the calls for various hosts are made is arbitrary; we cannot know which host will be pinged first. This results from the loop at https://github.com/python-diamond/Diamond/blob/master/src/collectors/ping/ping.py#L60, combined with the implementation details of `.keys()` (see https://docs.python.org/2/library/stdtypes.html#dict.items).

shortdudey123: fair point

Issue #525, opened 2016-08-31
Kafka Mbean with type 'java.lang.Object'


anshulp25:
`kafkastat.py` perfectly handles the mbeans having values type : `double`, `float`, `int` and `long`, but some mbeans ( for eg : `kafka.server:type=ReplicaManager,name=LeaderCount` ) has type `java.lang.Object`, which code ignores and doesn't extract values from it.

Any generic way to overcome this ?

josegonzalez:
Not sure, we'll definitely need a patch for that.

shortdudey123:
Solved by https://github.com/python-diamond/Diamond/pull/378

Issue #510, opened 2016-08-01
kafka per topic metrics (kafka >= 0.9)


adfel70:
hello,

the kafkastat collector for some reason supports only splitting 2 optional "=" in each mx4j metric at most.
this prevents getting the topic metrics (per each topic).

the fix is very small:

instead of:  

``` python
elif i > 0:
    key = objectname.split('=')[2]
```

should be something like:  

``` python
elif i > 0:
    key = objectname.split('=')[i+1]
```

this would be very helpful !

josegonzalez:
Can you make a pull request for this issue?

shortdudey123:
Solved by https://github.com/python-diamond/Diamond/pull/378

PR #561, opened 2016-12-15
[Kafka] Mbean with type 'java.lang.Object' & per topic metrics

hamelg:
Fix 2 issues :
https://github.com/python-diamond/Diamond/issues/525
https://github.com/python-diamond/Diamond/issues/510


shortdudey123:
Can you add testing for this? then we can make sure a regression does not occur in the future

shortdudey123:
looking good 👍

shortdudey123:

PR #562, opened 2016-12-15
Bind collector: Adding support for new some v3 XML bind9 stats and JSON v1

thardie:
Bind 9.11 has removed the old XML (v2) interface. This adds some support for key statistics under /server section of new v3 XML and v1 JSON interface.

Querying the root of the XML interface can cause bind to stop answering queries for several hundred milliseconds. Querying /server XML only 4-5ms, and /server for JSON 1-2ms.

### DIFF
--- a/src/collectors/bind/bind.py
+++ b/src/collectors/bind/bind.py
@@ -35,6 +35,10 @@ class BindCollector(diamond.collector.Collector):
             " - memory (Global memory usage)\n",
             'publish_view_bind': "",
             'publish_view_meta': "",
+            'data_format': "Bind stats version:\n" +
+            " - xml_v2 (Original bind stats version from 9.5)\n" +
+            " - xml_v3 (New xml version)\n" +
+            " - json_v1 (JSON replacement for XML)\n",
         })
         return config_help
 
@@ -63,6 +67,7 @@ class BindCollector(diamond.collector.Collector):
             # By default we don't publish these special views
             'publish_view_bind': False,
             'publish_view_meta': False,
+            'data_format': 'xml',
         })
         return config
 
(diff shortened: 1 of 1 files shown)


src/collectors/bind/bind.py
```diff
@@ -35,6 +35,10 @@ def get_default_config_help(self):
             " - memory (Global memory usage)\n",
             'publish_view_bind': "",
             'publish_view_meta': "",
+            'data_format': "Bind stats version:\n" +
+            " - xml_v2 (Original bind stats version from 9.5)\n" +
```
shortdudey123: You should pick `xml` or `xml_v2` and make sure its consistent throughout

shortdudey123:
Tests and doc need to be updated as well

shortdudey123:
pep8 errors need to be addressed as well

thardie:
Ooops. Still need to do tests and docs.

src/collectors/bind/bind.py
shortdudey123: Looks like this was corrected, thanks!

thardie:
I don't have a python 2.6 environment to test this in. What's your stance on this feature not working in python 2.6?

shortdudey123:
2.6 is still supported by the project technically, but i will let others speak to that point

docs/collectors/BindCollector.md
```diff
@@ -32,6 +32,12 @@ publish | resolver, server, zonemgmt, sockets, memory, | Available stats:<br>
  | list
```
shortdudey123: mode change?
`100644 → 100755`

docs/collectors/BindCollector.md
```diff
@@ -32,6 +32,12 @@ publish | resolver, server, zonemgmt, sockets, memory, | Available stats:<br>
...
 publish_view_bind | False |  | bool
 publish_view_meta | False |  | bool
+data_format | xml_v2 | Bind stats version:<br>
```
shortdudey123: The table gets messed up, not sure whats the best way to handle it though

PR #565, opened 2016-12-16
debian packaging update

yvlasenko:
* use dh-python instead of python-support
* add systemd service
* don't install /etc/init/diamond.upstart


josegonzalez:
This breaks the setup for non-systemd servers.

shortdudey123:
Taken care of by #710

PR #567, opened 2016-12-21
Fix infinite redirect loop with httpd collector, support HTTPS

alexmv:
Without these, a status page which redirects HTTP to HTTPS causes an infinite loop and wedges the local collector.


alexmv:
Will do!

shortdudey123:
Merging this anyway since the docs needs to be synched project wide and testing in general expanded.

PR #568, opened 2016-12-27
duplicate init files on rhel7 and rhel6

erezzarum:
On RHEL6 both upstart and init.d are configured
On RHEL7 both systemd and init.d are configured.
I also believe there is no need for using upstart on RHEL6, If you think there's a use for that, i'll fix my pull request.


erezzarum:
Late at night, didn't see i used an old repository, creating another pull request.
Sorry for that!

PR #570, opened 2016-12-29
add rpmspec option to not overwrite existing config files

hamelg:
Mark all files in /etc/diamond as configuration files in rpm package.
When updating, rpm will not overwrite theses files if have been modified.

### DIFF
--- a/rpm/install
+++ b/rpm/install
@@ -4,3 +4,9 @@
 %{!?python_sitelib: %global python_sitelib %(%{__python} -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")}
 %{!?python_sitearch: %global python_sitearch %(%{__python} -c "from distutils.sysconfig import get_python_lib; print(get_python_lib(1))")}
 %endif
+
+CONFIGFILES="
+%config(noreplace) /etc/diamond/*"
+echo "$CONFIGFILES" | cat INSTALLED_FILES - > INSTALLED_FILES.new
+mv INSTALLED_FILES.new INSTALLED_FILES
+


rpm/install
```diff
@@ -4,3 +4,9 @@
 %{!?python_sitelib: %global python_sitelib %(%{__python} -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")}
 %{!?python_sitearch: %global python_sitearch %(%{__python} -c "from distutils.sysconfig import get_python_lib; print(get_python_lib(1))")}
 %endif
+
+CONFIGFILES="
+%config(noreplace) /etc/diamond/*"
+echo "$CONFIGFILES" | cat INSTALLED_FILES - > INSTALLED_FILES.new
+mv INSTALLED_FILES.new INSTALLED_FILES
```
shortdudey123: What is the purpose of the `echo` and `mv`?

`%config(noreplace)` looks like it should be all you need since it will append files w/ `.rpmnew`

hamelg: rpm/install is used to build the %install section in rpm specfile. Then, the %install section is executed to build the list of files (populated by setup.py install ... --record=INSTALLED_FILES) inserted in the rpm package. This list is stored in the file INSTALLED_FILES : see the %files section in diamond.spec. So, we must add the %config(noreplace) statement in this file to identify the config files.
documentation : http://ftp.rpm.org/max-rpm/s1-rpm-inside-files-list-directives.html
To examine the build directory tree, you can add the -k option in Makefile : 
     58 buildrpm: sdist
     59         ./setup.py bdist_rpm -k \
     60                 --release=$(RELEASE) \
To check the config files in a rpm package  : 
rpm -qcp PACKAGE_FILE

shortdudey123: Gotcha, missed that part
Thanks!

PR #574, opened 2017-01-10
First implementation of springboot collector based on dropwizard

arnobroekhof:
First implementation of springboot collector based on dropwizard


shortdudey123:
needs doc file in `docs/collectors/`

PR #576, opened 2017-01-12
Fix minor bug in NetAppCollector doc string.

waynr:
I noticed this while looking at http://diamond.readthedocs.io/en/latest/collectors/NetAppCollector/


shortdudey123:
This needs to be fixed in the doc file as well

PR #579, opened 2017-01-18
Complete the list of accepted variables for phpfpm collector

gjedeer:


docs/collectors/PhpFpmCollector.md
```diff
@@ -15,6 +15,15 @@ A sample php-fpm config for this collector to work is
 pm.status_path = /fpm-status
 ```
 
+If the URL of the fpm-status page is http://127.0.0.1:8080/fpm-status, you need to set:
```
shortdudey123: This is not part of the docstring in `src/collectors/phpfpm/phpfpm.py` so it should be added there since this file is technically autogenerated as noted on lines 1 and 2

docs/collectors/PhpFpmCollector.md
```diff
@@ -25,6 +34,9 @@ pm.status_path = /fpm-status
 
 Setting | Default | Description | Type
 --------|---------|-------------|-----
+uri | fpm-status | Path part of the URL, with or without the leading / | string
```
shortdudey123: Alphabetize

gjedeer: What is this table generated from? I've found the above text in docstring but not the table.

shortdudey123: It comes from the k/v in `get_default_config_help` in [each collector](https://github.com/python-diamond/Diamond/blob/master/src/collectors/phpfpm/phpfpm.py#L33).  However, it looks like this was never filled in for the collector so this table only includes the [default ones](https://github.com/python-diamond/Diamond/blob/master/src/diamond/collector.py#L244)

Can you add the config options to that method? You can reference this [ntp](https://github.com/python-diamond/Diamond/blob/master/src/collectors/ntp/ntp.py#L34) collector if you need something to look at for it

gjedeer: Thanks! Done.

shortdudey123: cool, this list still needs to be alphabetized in this file though :)

gjedeer: Done. Also made lines in docstring shorter because pep8. One line is still >80 chars but it's a link, I don't know how to make it pep8-compatible without breaking the link.

It's backwards, the doc files should be automatically generated and build_doc.py doesn't work at all... Well I guess I have no right to complain if a patch isn't attached to the complaint.

shortdudey123:
@gjedeer Can you squash your commits to keep the history clean? then we can merge

gjedeer:
Can you merge this now? Rebased and squashed.

shortdudey123:
Can you rebase on top of master as opposed to pull master on top of? then we avoid extra merge commits that say `	Merge branch 'master' of https://github.com/python-diamond/Diamond ...`

gjedeer:
I don't know how to do that.

shortdudey123:
This is 2 good things to read about it:
https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request
http://rypress.com/tutorials/git/rebasing

Generally though
```
git checkout patch-1
git remote add upstream git@github.com:python-diamond/Diamond.git
git pull --rebase upstream master
```

Then it will tell you about any conflicts that exist (see them with `git status`), however, there should be none since there hasn't been any changes to the files you are changing.

gjedeer:
Thanks, so easy. Pushed the rebased version.

shortdudey123:
Awesome!

PR #581, opened 2017-01-20
Expand elasticsearch metrics on cluster health api

shortdudey123:
Elasticsearch 2+ adds several fields to the cluster health api (`/_cluster/health`) that are not reported via the existing collector code.


src/collectors/elasticsearch/test/testelasticsearch.py
```diff
@@ -119,6 +119,78 @@ def test_should_work_with_real_data(self, publish_mock):
...
+            'cluster_health.shards.active_percent': 100,
+            'cluster_health.shards.delayed_unassigned': 0,
+            'cluster_health.shards.relocating': 0,
+            'cluster_health.shards.unassigned': 0,
+            'cluster_health.shards.initializing': 0,
+
+            'cluster_health.status': 2,
+        }
+
+        self.setDocExample(collector=self.collector.__class__.__name__,
```
kormoc: Only one test should run setDocExample, as only the first executed test will run the placeholder replacement.

shortdudey123: Ah true, thanks for catching that

shortdudey123: fixed

PR #585, opened 2017-01-24
Adding more information on how to set the metric path on a collector, fixing #584

### LINKED ISSUE #584: How to set my metric path when creating a custom collector?
Hi

I'm trying to create my own collector, but I'm having problems with the path. The collector is called "MyCustomCollector" and the code is as simple as:
```
import diamond.collector

class MyCustomCollector(diamond.collector.Collector):
    def collect(self):
        metric_name = "my_test"
        metric_value = 42
        self.publish(metric_name, metric_value)
```
I expected to find the metric at servers.hostname.my_test, but it's being written to servers.hostname.MyCustomCollector.my_test. There's nothing about setting the metric path on the ExampleCollector documentation, so I'm not sure how to change that.

Any hint?

Thanks!

ersiko:
I was trying to set up a new collector using ExampleCollector as an example (duh!) but I didn't know how to set the metric path. After asking in issue #584 I got the information and I created this PR so it would be available for anyone else trying to create their own collectors. Fixes #584

### DIFF
--- a/src/collectors/example/example.py
+++ b/src/collectors/example/example.py
@@ -21,6 +21,13 @@ form, they need to implement a single method called "collect".
     class ExampleCollector(diamond.collector.Collector):
 
         def collect(self):
+        
+            # Set Metric Path. By default it will be the collector's name
+            # (servers.hostname.ExampleCollector.my.example.metric)
+            self.config.update({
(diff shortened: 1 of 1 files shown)


kormoc:
Thanks!

PR #609, opened 2017-02-19
Add linux raid (mdstat) collector

### LINKED ISSUE #608: Feature: Add a collector for linux raid
I propose to add a collector for linux raid.  
Linux raid can be monitored by parsing `/proc/mdstat`: https://raid.wiki.kernel.org/index.php/Mdstat

nightvisi0n:
This PR adds a new collector for metric gathering of linux raid devices.  
It reads `/proc/mdstat` and has no other dependencies.  
I already tested it on production with `NullHandler` - :+1: works for me.

fixes #608

### DIFF
--- a/src/collectors/mdstat/mdstat.py
+++ b/src/collectors/mdstat/mdstat.py
@@ -2,12 +2,48 @@
 
 """
 Collect linux RAID/md state by parsing /proc/mdstat.
-https://raid.wiki.kernel.org/index.php/Mdstat
+<https://raid.wiki.kernel.org/index.php/Mdstat>
 
 #### Dependencies
 
- * /proc/mdstat
+- /proc/mdstat
 
+#### Supported metrics
+
+```
+md0 : active raid1 sda1[0] sda2[2](S) sda3[1]
+```
+- member_count.active
+- member_count.faulty
+- member_count.spare
+
+```
+39058432 blocks super 1.2 level 5, 512k chunk, algorithm 2 [3/3] [UUU]
+199800 blocks super 1.2 999k rounding
+```
+- status.blocks
+- status.superblock_version
+- status.raid_level
+- status.chunk_size
+- status.algorithm
+- status.rounding_factor
+- status.actual_members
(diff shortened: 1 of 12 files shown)


src/collectors/mdstat/mdstat.py
```diff
@@ -0,0 +1,317 @@
...
+                'Error opening {mdstat_path} for reading: {err}'.format(
+                    mdstat_path=self.MDSTAT_PATH,
+                    err=err
+                )
+            )
+
+        arrays = {}
+        mdstat_array_blocks = ''
+
+        for line in lines[1:-1]:
```
shortdudey123: If the IOError is caught, then `lines` is never defined and this line will cause a `NameError`

src/collectors/mdstat/test/test_mdstat.py
```diff
@@ -0,0 +1,230 @@
...
+            'md0.member_count.active': 2,
+            'md0.member_count.faulty': 0,
+            'md0.member_count.spare': 0,
+            'md0.bitmap.total_pages': 1,
+            'md0.bitmap.allocated_pages': 1,
+            'md0.bitmap.page_size': 4,
+            'md0.bitmap.chunk_size': 65536
+        }
+
+        self.setDocExample(collector=self.collector.__class__.__name__,
```
shortdudey123: `self.setDocExample`  should only be called once per test file, so one of the calls needs to be removed

nightvisi0n: thanks, completely overlooked this one :see_no_evil:  
fixed with e5ff7b1

src/collectors/mdstat/mdstat.py
nightvisi0n: fixed with 849e738

nightvisi0n:
@shortdudey123 Thanks for the review.  
I provided the requested changes and also did some refactoring and documentation improvements.  
Can you please review again? :slightly_smiling_face:

PR #611, opened 2017-02-22
Fix diamond not shutting down gracefully 595

DStape:


shortdudey123:
Awesome! thanks for getting this fixed

PR #612, opened 2017-02-28
Add support for Aerospike v3.9+ metric format

travisgroth:
Previous version worked with Aerospike through v3.9, when some metrics changed names and/or were reformatted.  

Updates unit testing, whitelists, and does some version checking to correctly parse statistics and latency data.  Should be easy to add another "dialect" in the future now.


travisgroth:
@shortdudey123 is there a way to kick coveralls in the head?  Would like to get feedback on this and get it merged quickly.  We need to push this update out internally and we're trying to avoid getting too far out of sync with upstream.

src/collectors/aerospike/aerospike.py
```diff
@@ -78,32 +87,82 @@ def get_default_config(self):
             'migrate-rx-partitions-initial',
             'migrate-rx-partitions-remaining',
             'available_pct',
+            # 3.9 Stats
+            "client_delete_error",
```
shortdudey123: double quotes?

src/collectors/aerospike/aerospike.py
```diff
@@ -61,8 +63,15 @@ def get_default_config(self):
             'objects',
             'client_connections',
             'index-used-bytes-memory',
+            # 3.9 Stats
+            'objects',
```
shortdudey123: some things are listed twice, maybe group them under `# 2.7 and 3.9 Stats` or `# All version stats`

shortdudey123: Same w/ `namespace_statistics_whitelist`

travisgroth: Yeah was aware of that.  Though it's slightly less efficient, I thought it was clearer/easier to show and track exactly what was deemed useful out of the box in each version since they're obviously subject to change again in the next dialect change.  In fact, I'm nearly positive they'll get rid of the -'s in the near future.  If you have a scenario where something is in v2.7 and 3.9 but not v3.14, you're back to duplicating or making it more complicated.

I could throw them in a set to offset the cost of the 'in' tests and keep the maintainability and clarity.  Thoughts?

src/collectors/aerospike/aerospike.py
travisgroth: Fixed.

src/collectors/aerospike/aerospike.py
shortdudey123: hmm true
Probably fine for now

PR #614, opened 2017-03-06
Fixed simple bug "in" vs "=="

jwasher:
Was reading Diamond as it's pointed to as an example of good Python style. Oddly enough, I think I spotted a bug, using '==' instead of 'in' with a list

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -84,7 +84,7 @@ else:
     if running_under_virtualenv():
         install_requires = ['configobj', 'psutil', ]
     else:
-        if distro in ['debian', 'ubuntu']:
+        if distro in ['debian', 'Ubuntu']:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -84,7 +84,7 @@ def running_under_virtualenv():
     if running_under_virtualenv():
         install_requires = ['configobj', 'psutil', ]
     else:
-        if distro == ['debian', 'ubuntu']:
+        if distro in ['debian', 'ubuntu']:
```
shortdudey123: Looks like this only partially fixes the bug.  `ubuntu` should be capitalized

```
>>> import platform
>>> platform.dist()[0]
'Ubuntu'
>>>
```

PR #615, opened 2017-03-11
Python3 support

shortdudey123:
Rebased version of https://github.com/python-diamond/Diamond/pull/259:

This PR makes it possible to run and build Diamond with Python3.

Note that this breaks compatility with Pythons before 2.6 but that has mostly been the de facto state for a lot of the collectors even before this. (they depend on things like subprocess.check_output)


grobian:
this needs in addition:
- Queue -> queue in src/diamond/handler/queue.py for import Queue and Queue.Full
- removal of paths = map(str.strip, paths) from src/diamond/utils/classes.py (to load anything)
- use of str.encode(data) with self.socket.sendall in src/diamond/handler/graphite.py

I don't know how to add those to this PR without creating a new one.  I'd be cool if we could get some traction on merging Python3 compatibility.

vladislavkoz:
Hello guys. Any updates on this PR? Do you have some information about Python3 support?

myroslavsok:
Oh, I need Python 3 support as well. It would be great.

mpcathey:
Is this the best place to revive the python3 support discussion?

vladislavkoz:
> Is this the best place to revive the python3 support discussion?

Just check the number of forks. I believe everybody just doing it's own fork now and upgrading to p3

PR #617, opened 2017-03-17
Fixes the missing metrics from kafka.Controller:type=KafkaController

prasincs:
All the metrics under `kafka.Controller:type=KafkaController` were missing because Kafka reports the stats as `java.lang.Object`.These are critical metrics like OfflinePartitionCount and
ActiveControllerCount.

This patch also handles if the mbeans return empty or None for some reason. It will log for that particular objectName and move on. This will attempt to parse more objects than previous version but it's working okay for us. If it fails to find a numeric type, it will exit creating metrics. Since there was no fixture for this case I have added that too.


prasincs:
This has been fixed, I think it's somewhat of a bug fix rather than enhancement because some critical Kafka metrics were missing because the collector was ignoring them because of the type. I've added tests to fix these issues in future and to catch any regressions.

PR #619, opened 2017-03-22
Add role stat to redisstat collector

shortdudey123:
If the redis instance is a master, then the `replication.master` metric will be 1.
Otherwise it will be 0 (assuming slave).

### DIFF
--- a/src/collectors/redisstat/redisstat.py
+++ b/src/collectors/redisstat/redisstat.py
@@ -101,6 +101,7 @@ class RedisCollector(diamond.collector.Collector):
              'process.uptime': 'uptime_in_seconds',
              'pubsub.channels': 'pubsub_channels',
              'pubsub.patterns': 'pubsub_patterns',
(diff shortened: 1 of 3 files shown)


src/collectors/redisstat/redisstat.py
```diff
@@ -303,6 +304,19 @@ def collect_instance(self, nick, host, port, unix_socket, auth):
         # server
         data = dict()
 
+        # Role needs to be handled outside the the _KEYS dict
+        # since the value is a string, not a int / float
+        # Also, master_sync_in_progress is only available if the
+        # redis instance is a slave, so default it here so that
+        # the metric is cleared if the instance flips from slave
+        # to master
+        if 'role' in info.keys():
```
WeiBanjo: I think you can just do `'role' in info`

shortdudey123: ah, true

```
>>> data = { 'role': 'master' }
>>> data
{'role': 'master'}
>>> 'role' in data
True
>>> 
```

PR #621, opened 2017-03-24
More variables in httpd collector

gjedeer:
The Httpd collector had a very limited list of collected variables, probably based on someone's particular needs. I've added all values reported by Apache 2.4.18 with mpm_event - it should be the user's choice to exclude some variables, not the developer's.


src/collectors/httpd/httpd.py
```diff
@@ -138,6 +151,12 @@ def _publish(self, nickname, key, value):
             if len(nickname) > 0:
                 metric_name = nickname + '.' + metric_name
 
+            # Strip percent mark from Cache*Usage
+            try:
```
shortdudey123: Can you update the tests to test this logic?

gjedeer: done

PR #622, opened 2017-03-25
set precision of TCPCollector's metrics to 2

ikruglov:
Having precision=0 means that any non frequent metrics will be hidden by
diamond. For example: if interval=60 and a counter incremented once a minute
we have following math done by derivate():

old=1
new=2
derivative_x = 2-1 = 1
derivative_y = 60
result = 1 / 60 = 0.01666666666667

this value is later passed to Metric class which does following if
precision=0:

value = sprintf("%.0f", 0.01666666666667)

as result value becomes 0.

### DIFF
--- a/src/collectors/tcp/tcp.py
+++ b/src/collectors/tcp/tcp.py
@@ -272,6 +272,6 @@ class TCPCollector(diamond.collector.Collector):
 
             # Publish the metric
             if metric_name in self.config['gauges']:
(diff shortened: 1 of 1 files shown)


shortdudey123 (changes requested):
Can you make this a config option instead of changing the hard coded value?

shortdudey123:
Also update `docs/collectors/TCPCollector.md` with the `precision` config option

PR #626, opened 2017-03-30
use dict(Counter(x)+Counter(y)) to calc sum of dicts

williamjoy:
fix syntax error on python 2.6

```
[2017-03-30 06:56:15,401] [MainThread] Failed to import module: mesos. Traceback (most recent call last):
  File "/usr/lib/python2.6/site-packages/diamond/utils/classes.py", line 159, in load_collectors_from_paths
    mod = imp.load_module(modname, fp, pathname, description)
  File "/usr/share/diamond/collectors/mesos/mesos.py", line 175
    for key in stats
      ^
SyntaxError: invalid syntax
```

### DIFF
--- a/docs/collectors/MesosCollector.md
+++ b/docs/collectors/MesosCollector.md
@@ -35,6 +35,17 @@ port | 5050 | Port (default is 5050; set to 5051 for mesos-agent) | int
 #### Example Output
 
 ```
-__EXAMPLESHERE__
+servers.hostname.mesos.failed_tasks 6
+servers.hostname.mesos.finished_tasks 1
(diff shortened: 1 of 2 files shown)


shortdudey123:
Based on the fact that the bug was not causing tests to fail, can you add a test to cover the method that you fixed? that will keep it from being broken in the future.

williamjoy:
The code is incompatible with python 2.6, and `collection.Counter` also does not work for python 2.6. 

Closing this and looking for other solution

shortdudey123:
Gotcha, thanks for looking at this :)
Can you file an issue as well? that way it can be tracked if you don't have a chance to figure out another solution

PR #627, opened 2017-03-30
Collect test coverage results for collectors too

shortdudey123:


shortdudey123:
Looks like each collector also needs `__init__.py` to be considered by `coverage`

`Only importable files (ones at the root of the tree, or in directories with a __init__.py file) will be considered.` ~ http://coverage.readthedocs.io/en/coverage-4.3.4/source.html#execution

```
$ find src -name __init__.py
src/diamond/__init__.py
src/diamond/handler/__init__.py
src/diamond/utils/__init__.py
$ 
```

PR #629, opened 2017-04-01
dict generator does not work for python 2.6

williamjoy:
duplicated with pull request #626

https://www.python.org/dev/peps/pep-0274/#semantics

```
The semantics of dict comprehensions can actually be demonstrated in stock Python 2.2, by passing a list comprehension to the built-in dictionary constructor:

>>> dict([(i, chr(65+i)) for i in range(4)])
is semantically equivalent to:

>>> {i : chr(65+i) for i in range(4)}
```

### DIFF
--- a/src/collectors/mesos/test/testmesos.py
+++ b/src/collectors/mesos/test/testmesos.py
@@ -2,7 +2,6 @@
 # coding=utf-8
 ##########################################################################
 
-import json
 from test import CollectorTestCase
 from test import get_collector_config
 from test import unittest
@@ -166,10 +165,11 @@ class TestMesosCollector(CollectorTestCase):
                          self.collector._get_url("metrics/snapshot"))
 
     def test_sum_statistics(self):
-        results = json.load(self.getFixture('slave_monitor_statistics.json'))
-        sum = {}
-        for i in results:
-            sum = self.collector._sum_statistics(sum, i['statistics'])
+        metrics_1 = {'cpu': 50, 'mem': 30, 'loadavg': 1}
+        metrics_2 = {'cpu': 10, 'mem': 30, 'network': 10}
+        self.assertEqual(self.collector._sum_statistics(metrics_1, metrics_2),
+                         {'mem': 60, 'loadavg': 1, 'network': 10, 'cpu': 60})
+
 
 ##########################################################################
 if __name__ == "__main__":
--- a/src/collectors/mesos/test/testmesos.py
+++ b/src/collectors/mesos/test/testmesos.py
@@ -2,6 +2,7 @@
 # coding=utf-8
 ##########################################################################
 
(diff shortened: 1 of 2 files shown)


shortdudey123:
Since the bug was not causing tests to fail, can you add a test to cover the `_sum_statistics ` method? That will keep it from getting broken in the future

williamjoy:
hi @shortdudey123 , I tried to add a test case which might cover it

test against old mesos.py
```
$ python2.7 ./test.py -c mesos -v
test_http (testmesos.TestMesosCollector) ... ok
test_https (testmesos.TestMesosCollector) ... ok
test_import (testmesos.TestMesosCollector) ... ok
test_should_compute_cpus_percent (testmesos.TestMesosCollector) ... ok
test_should_compute_cpus_utilisation (testmesos.TestMesosCollector) ... ok
test_should_fail_gracefully (testmesos.TestMesosCollector) ... ok
test_should_work_for_master_with_real_data (testmesos.TestMesosCollector) ... ok
test_should_work_for_slave_with_real_data (testmesos.TestMesosCollector) ... ok
test_sum_statistics (testmesos.TestMesosCollector) ... ok

----------------------------------------------------------------------
Ran 9 tests in 0.014s

OK
$  python2.6 ./test.py -c mesos -v
Failed to import module: testmesos. Traceback (most recent call last):
  File "./test.py", line 246, in getCollectorTests
    ['*'])
  File "/home/william.wei/Diamond/src/collectors/mesos/test/testmesos.py", line 14, in <module>
    from mesos import MesosCollector
  File "/home/william.wei/Diamond/src/collectors/mesos/mesos.py", line 177
    for key in stats
      ^
SyntaxError: invalid syntax


----------------------------------------------------------------------
Ran 0 tests in 0.000s

OK
```

src/collectors/mesos/test/testmesos.py
```diff
@@ -164,6 +165,12 @@ def test_https(self):
         self.assertEqual('https://localhost:5050/metrics/snapshot',
                          self.collector._get_url("metrics/snapshot"))
 
+    def test_sum_statistics(self):
+        results = json.load(self.getFixture('slave_monitor_statistics.json'))
+        sum = {}
+        for i in results:
+            sum = self.collector._sum_statistics(sum,i['statistics'])
```
shortdudey123: can you add `self.assertEqual(XXX, sum)`? (XXX being the sum that should be there)

shortdudey123:
awesome! thanks for showing the output of the new test with the old code :)

src/collectors/mesos/test/testmesos.py
williamjoy: sure, I am just learning how to write test cases, thanks for your guide :)

PR #635, opened 2017-04-24
Add PostgreSQL handler

cgalisteo:
Adding a PostgreSQL handler based on the already available MySQL one.

### DIFF
--- a/src/diamond/handler/postgresql.py
+++ b/src/diamond/handler/postgresql.py
@@ -1,0 +1,110 @@
+# coding=utf-8
+
+"""
+Insert the collected values into a  table
+"""
+
+from Handler import Handler
+import psycopg2
+
+
+class PostgreSQLHandler(Handler):
+    """
+    Implements the abstract Handler class, sending data to a mysql table
+    """
+    conn = None
+
+    def __init__(self, config=None):
+        """
+        Create a new instance of the PostgreSQLHandler class
+        """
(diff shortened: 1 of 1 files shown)


src/diamond/handler/postgresql.py
```diff
@@ -0,0 +1,110 @@
+# coding=utf-8
+
+"""
+Insert the collected values into a  table
```
shortdudey123, line 4: Replace double space near end with postgreSQL

src/diamond/handler/postgresql.py
```diff
@@ -0,0 +1,110 @@
...
+"""
+
+from Handler import Handler
+import psycopg2
+
+
+class PostgreSQLHandler(Handler):
+    """
+    Implements the abstract Handler class, sending data to a mysql table
```
shortdudey123, line 13: mysql should be postgresql

src/diamond/handler/postgresql.py
```diff
@@ -0,0 +1,110 @@
...
+        self._connect()
+
+    def get_default_config_help(self):
+        """
+        Returns the help text for the configuration options for this handler
+        """
+        config = super(PostgreSQLHandler, self).get_default_config_help()
+
+        config.update({
+        })
```
shortdudey123, line 45: Can you add descriptions for all the config options?

PR #616, opened 2017-03-15
TSDB basic authorization, gzip, batch, prefix

Grot4x:
Added some features to the TSDB Handler
basic authorization: simple header with user and password for firewall protection
gzip: added gzip support to compress metrics
batch: added support to send metrics in batch
prefix: you can add a prefix to all you metrics like diamond.myhostname.cpu.cpu_count

All these features can be disabled by not defining or setting a value lower than 1.

Sending metrics in batch can give you quite a nice performance boost and lower you cpu load.
Compressing will reduce the buffer size you need to configure on the tsdb end.

On the negative side this will break the recently added tests and I was not able to recreate them.
I would need some help as there is currently no handler (using urllib2) with tests.
But I didn't want to keep these changes for my self.
batch
This graph shows you how long it took to send a metric to the db.

PR #637, opened 2017-04-26
Fix tsdb metric names when cleaning

DarrellMozingo:
PR #616 made a regression where the cleaned metric names weren't getting sent to TSDB (it was building the metric before cleaning the metric). Fixed that and added a few extra assertions around compression.


shortdudey123:
Awesome, thanks!

PR #639, opened 2017-04-28
patch for fixing systemd service restart problem

sideffect0:
This fixes start problem when using systemctl restart

### DIFF
--- a/rpm/systemd/diamond.service
+++ b/rpm/systemd/diamond.service
@@ -2,6 +2,7 @@
 Description=diamond - A system statistics collector for graphite
 
 [Service]
+ExecStartPost=/bin/sleep 0.1
 ExecStart=/usr/bin/python /usr/bin/diamond
 Type=forking
 #If not using default pid, make sure you change PIDFile setting


shortdudey123:
Can you link to which specific issue you are solving for reference? or provide a lot more context on what you are seeing?

sideffect0:
On `restart` ing the service with  
`systemctl restart diamond`  
The command will fail with error because,
`diamond.service: Start request repeated too quickly.`  
with recent version of systemd in ubuntu or debian distros.

shortdudey123:
Can you create an issue with full info to go along with this PR then?  I have yet to see that message and i use centos7 exclusively

sideffect0:
https://github.com/python-diamond/Diamond/issues/640

trinitronx:
Note: using `sleep` can introduce race conditions here... especially with SystemD which likes to spawn as many services in parallel as possible.

If any other service uses `Wants=diamond.service`, or `Requires=diamond.service`, expecting that Diamond is running... this guarantee can not be met and you will encounter problems with race conditions on service startup ordering during system boot.

Better to use `Type=notify`, and patch Diamond to work directly with SystemD and play nicely with it.  That will avoid all types of race condition problems!

shortdudey123:
See #640

PR #644, opened 2017-05-02
Run lsof with sudo, so that the reported numbers are not 0

NotSqrt:
Hi !

I've discovered that the numbers reported by filestat were always 0, expect for the user running diamond.
So typically a problem of permissions.

My solution: add `use_sudo` option, and the `sudoers` config to make it work without passwords.

I faced this issue on a ubuntu 12.04 machine.

Thanks !


src/collectors/filestat/filestat.py
```diff
@@ -106,7 +118,9 @@ def get_default_config(self):
             'uid_max': 65536,
             'type_include': None,
             'type_exclude': None,
-            'collect_user_data': False
+            'collect_user_data': False,
+            'use_sudo': False,
+            'sudo_cmd': '/usr/bin/sudo',
```
shortdudey123: these should be added to the `get_default_config_help` section too

NotSqrt: Added.

src/collectors/filestat/filestat.py
```diff
@@ -233,28 +309,20 @@ def process_lsof(self, users, types):
...
         """
-        d = {}
-        for u in users:
-            d[u] = {}
-            tmp = os.popen("lsof -wbu %s | awk '{ print $5 }'" % (
-                u)).read().split()
-            for t in types:
-                d[u][t] = tmp.count(t)
-        return d
+        return parse_lsof_output(self.get_output('tL'), users, types)
```
NotSqrt: Switched to `lsof -F tL`, because I found that the type is not always in `awk '{ print $5 }'`.

src/collectors/filestat/filestat.py
```diff
@@ -106,10 +160,24 @@ def get_default_config(self):
...
-            'collect_user_data': False
+            'collect_user_data': False,
+            'use_sudo': False,
+            'sudo_cmd': '/usr/bin/sudo',
         })
         return config
 
+    def get_output(self, fmt):
+
+        command = ['lsof', '-F', fmt]
```
shortdudey123, line 171: Can you use the full path for lsof?

NotSqrt: This PR is old :) I'm no longer using diamond..

Will it be portable to hardcode `/usr/bin/lsof` ?

PR #646, opened 2017-05-09
Add multihost support for PostgreSQL collector

wstranger:


src/collectors/postgres/postgres.py
```diff
@@ -79,53 +102,73 @@ def collect(self):
...
-            metrics = self.config['metrics']
-        elif str_to_bool(self.config['extended']):
-            metrics = registry['extended']
-            if str_to_bool(self.config['has_admin']) \
-                    and 'WalSegmentStats' not in metrics:
-                metrics.append('WalSegmentStats')
+        for instance in self.config['instances'].keys():
+            # Get list of databases
+            #import pdb
+            #pdb.set_trace()
```
shortdudey123: probably can remove this

src/collectors/postgres/postgres.py
```diff
@@ -54,9 +76,9 @@ def get_default_config(self):
         Return default config.
         """
         config = super(PostgresqlCollector, self).get_default_config()
+
         config.update({
             'path': 'postgres',
-            'host': 'localhost',
```
shortdudey123: Currently this PR is a breaking change.  However, if you keep this config option, add it to the instance list in the `collect` method if it isn't already in it, and then exclude `host` from falling back to the default in `_get_config`, then it would not be a breaking change.

wstranger: I already added it in my last commit in the `collect` method. Same way as pgbouncer collector.

shortdudey123: I don't see the `host` config option being used in https://github.com/python-diamond/Diamond/pull/646/commits/e6d03ba7e67b91cf0e2787d7481e9b2a40876a06

kormoc:
Can you quickly give a run down on why the multiple collector core code doesn't work for this? Thanks!

wstranger:
I didn't find any suitable implementation in core and use pgbouncer collector code as reference.

kormoc:
Hrm. I can't find it documented in the docs :(

Basically, core allows you to spin up multiple instances of any collector via multiple config files of the pattern 
`/etc/diamond/collectors/ExampleCollector - foobar.conf`. Every individual config file will be a entirely independent collector process and can be configured as such.

wstranger:
Ok. At the doc http://diamond.readthedocs.io/en/latest/Getting-Started/Configuration/ - we can see next - 
> The configuration file name should match the name of the diamond collector class.
Hrm. Nothing about wildcards.
Slony, Redis, Elastic, pgbouncer collectors use same way as this one.
We are planning to use diamond in docker environment and I'm afraid it is not convenient to  manage over 100 configuration files.
Thank you.

shortdudey123:
PR LGTM now
However, i will defer to @kormoc on merging or not

src/collectors/postgres/postgres.py
```diff
@@ -79,53 +103,81 @@ def collect(self):
...
-        elif str_to_bool(self.config['extended']):
-            metrics = registry['extended']
-            if str_to_bool(self.config['has_admin']) \
-                    and 'WalSegmentStats' not in metrics:
-                metrics.append('WalSegmentStats')
+        instances = self.config.get('instances')
+
+        # HACK: setting default with subcategory messes up merging of configs,
+        # so we only set the default if one wasn't provided.
+        if not instances:
```
shortdudey123, line 110: Won't this code never run since you defined a default for instances that is not None?

src/collectors/postgres/postgres.py
```diff
@@ -79,53 +103,81 @@ def collect(self):
...
+            instances = {
```
shortdudey123, line 111: You need to set to self.instances since you use it in _get_fig()

src/collectors/postgres/test/testpostgres.py
```diff
@@ -14,43 +14,60 @@ class TestPostgresqlCollector(CollectorTestCase):
     def setUp(self, allowed_names=None):
         if not allowed_names:
             allowed_names = []
+
+        default_config = get_collector_config('PostgresqlCollector', {})
+        self.default_collector = PostgresqlCollector(default_config, None)
```
shortdudey123, line 19: Doesn't look like you ever used this.  Can you add tests using it to test the default config behavior?

PR #652, opened 2017-05-16
Include xa-datasource statistics to app metrics

skythet:
Current jbossapi collector fetch only data-sources skipping xa-data-sources, in this commit I included xa-data-sources for metrics

### DIFF
--- a/src/collectors/jbossapi/jbossapi.py
+++ b/src/collectors/jbossapi/jbossapi.py
@@ -208,6 +208,9 @@ class JbossApiCollector(diamond.collector.Collector):
             if op_type == 'app' and self.config['app_stats'] == 'True':
                 if output:
                     # Grab the pool stats for each Instance
+                    if 'xa-data-source' in output['result']:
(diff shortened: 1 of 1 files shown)


src/collectors/jbossapi/jbossapi.py
```diff
@@ -387,7 +389,7 @@ def collect(self):
             current_proto = matches.group(5)
             current_user = matches.group(1)
             current_pword = matches.group(2)
-
+
```
shortdudey123: revert

src/collectors/jbossapi/jbossapi.py
```diff
@@ -374,7 +377,6 @@ def string_regex(self, pattern, s):
         return tmp_result.group(1)
 
     def collect(self):
-
```
shortdudey123: revert

shortdudey123:
Still needs tests added for the metric
Travis says `abuse detected: known offender (request looked fishy)` so i will hold off merging til that is resolved as well

PR #654, opened 2017-06-02
fix undefined var 'user'

joelrebel:
when a user, group is defined in diamond.conf, we end up with the below error,

 ERROR: Failed to set UID/GID. global name 'user' is not defined

Seems to be introduced in [ad3862d76c](https://github.com/python-diamond/Diamond/commit/ad3862d76c583f9ec623758344165da11cef8c62 ),

This fixes it on nix based OS's, it  needs testing on 'nt' based systems since 'pwd' might not be available on those

### DIFF
--- a/bin/diamond
+++ b/bin/diamond
@@ -201,6 +201,7 @@ def main():
             try:
                 if gid != -1 and uid != -1:
                     # Manually set the groups since they aren't set by default
+                    user = pwd.getpwuid(uid).pw_name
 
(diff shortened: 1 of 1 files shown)


JakSla:
this can also be used as an alternative
user = config['server']['user']

shortdudey123:
Can you rebase on master? As a side note, the alternative mentioned above won't work since user has two config sources, not one.

PR #658, opened 2017-06-28
correct execute permission checking in nagios collector

wleese:
The current boolean with cause the collect() to return when you enable sudo, but as the current user cannot access the nagiosstats binary. This seems incorrect because that is precisely the reason one would use sudo.


wleese:
I believe I corrected the whitespace, but travis needs to be re-triggered?

PR #665, opened 2017-07-28
Allow setting a timestamp when publishing a metric

TomTheBear:
I have collectors that publish multiple metrics in one `process` run. Since it takes some time to get the data, I would like to set the timestamp when calling the `publish` function. This PR just adds the timestamp as a optional parameter to the `publish` function and forwards the value to the Metric class.


shortdudey123:
Used #448 instead.

PR #504, opened 2016-07-07
Update InfluxDBHandler for post InfluxDB 0.9

cj-dimaggio:
- Fix for ticket #297 where InfluxDBHandler was formatting metrics based
  on an older version of influxdb-python
- Maintains support for InfluxDB version 0.8 by way of an additional
  'influxdb_version' attribute in the config.
- Reformats measurement schema to be more useful for InfluxDB 0.9's removal
  of merge and joins.
- Adds several test cases for the handler.

PR #666, opened 2017-07-28
Use valid format for InfluxDBClient.write_points and convert metric path to metric name plus tags

TomTheBear:
In my tests, the InfluxDB handler didn't work because the JSON dict created by the handler wasn't accepted by python-influxdb. I re-checked the implementation in python-influxdb and the problem is definitely on at the handler side.
This PR includes:
- Proper dict keys for python-influxdb
- Transform the metric path to metric name plus tags. This is probably not complete, currently only hostname and if available CPU are added to tags. The metric name will be <collector>_<name>
- Remove the timestamp test for sending a batch. This could cause a loss of metrics when the batch buffer is almost full and the handler should process a bunch of metrics coming in in a small time range.

### DIFF
--- a/src/diamond/handler/influxdbHandler.py
+++ b/src/diamond/handler/influxdbHandler.py
@@ -169,20 +169,17 @@ class InfluxdbHandler(Handler):
                 metrics = []
                 for path in self.batch:
                     pathlist = path.split(".")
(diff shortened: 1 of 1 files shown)


TomTheBear:
There is already the PR #504 doing the same stuff

Issue #638, opened 2017-04-28
python-2.4 is not working


bhepple:
The docs claim that python-2.4 is supported (eg on RHEL5) but diamond fails with:

ERROR: Failed to set UID/GID. 'module' object has no attribute 'initgroups'

/usr/bin/diamond:204: os.initgroups(pwd.getpwuid(uid).pw_name, gid)

shortdudey123:
We haven't tested against 2.5 in quite a while
2.5 testing was removed in https://github.com/python-diamond/Diamond/commit/a08bc489751b2580f9e2e690c26efc20dd1df0c0 over 3 yrs ago and 2.4 testing was never done in travis.  If 2.4 or 2.5 is broken, then we should probably drop support as opposed to fixing.

That being said, @bhepple you are free to help work on fixing it.  However, Travis no longer supports python <2.6 so it getting broken again it fairly likely.

bhepple:
python26 would be fine for RHEL5 - but the python27-isms kill that too

shortdudey123:
FWIW RHEL 5 is EOL so we won't support it
which `python27-isms` don't work on RHEL 6/7?

bhepple:
Please see original post - initgroups() is the first I ran into. I should have mentioned that this is the latest release Diamond-4.0.515

Cheers


Bob

shortdudey123:
I am able to start diamond master on CentOS 6 with no issues

shortdudey123:
Added the vagrant box i used for testing via https://github.com/python-diamond/Diamond/pull/643
Can you give me more info on how to replicate?

bhepple:
Are you running python26? That's the default on RHEL6.

This snippet is isolated from line 204 of /usr/bin/diamond and it fails on RHEL6:

```
#!/usr/bin/env python
# coding=utf-8

import os
uid = 806 # our diamond user
gid = 751 # our diamond group
os.initgroups(pwd.getpwuid(uid).pw_name, gid)

$ python -V
Python 2.6.6

$ ./d
Traceback (most recent call last):
  File "./d", line 7, in <module>
    os.initgroups(pwd.getpwuid(uid).pw_name, gid)
AttributeError: 'module' object has no attribute 'initgroups'
```

https://docs.python.org/2/library/os.html states that:
```
os.initgroups(username, gid)
Call the system initgroups() to initialize the group access list with all of the groups of which the specified username is a member, plus the specified group id.

Availability: Unix.

New in version 2.7.
```

[...]

PR #676, opened 2017-08-28
Fix regression introduced by #638

olivierHa:
Hello 

Fixing #638 has introduced a regression with commit ad3862d76c583f9ec623758344165da11cef8c62
Indeed, user isn't defined.

Regards


JakSla:
this can also be used as an alternative
user = config['server']['user']

shortdudey123:
Can you rebase on master? As a side note, the alternative mentioned above won't work since user has two config sources, not one.

PR #679, opened 2017-10-06
CloudWatch custom dimensions and boto3

### LINKED ISSUE #678: CloudWatch handler uses deprecated SDK
The CloudWatch handler uses the deprecated [boto](http://boto.readthedocs.io/en/latest/) library instead of the current [boto3](https://boto3.readthedocs.io/en/latest/) / [botocore](https://botocore.readthedocs.io/en/latest/) library.

This means it's impossible to run Diamond in an ECS (docker) task using a Task Role, as those creds aren't supported by the old ``boto`` library.

jantman:
This rolls up two improvements for the cloudwatch handler:

1. Fix #678, switch from the deprecated boto AWS library (which, in addition to being deprecated, also doesn't support retrieving AWS credentials from the Instance Metadata Service or ECS Task Metadata).
2. Add support for custom dimensions to send along with metrics, per-metric.

### DIFF
--- a/src/diamond/handler/cloudwatch.py
+++ b/src/diamond/handler/cloudwatch.py
@@ -6,7 +6,7 @@ Automatically adds the InstanceId Dimension
 
 #### Dependencies
 
- * [boto](http://boto.readthedocs.org/en/latest/index.html)
+ * [boto3](http://boto3.readthedocs.org/en/latest/)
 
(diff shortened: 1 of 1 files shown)


shortdudey123:
#707 added custom dimension support.  Can you rebase this to get the boto3 switch done?

PR #688, opened 2017-12-07
fix TypeError: super(type, obj): obj must be an instance or subtype issue

jamie-liu:
When run diamond with ceph, collector  CephStatsCollector is failed to load due to "TypeError: super(type, obj): obj must be an instance or subtype of type".
According to the [link](https://thingspython.wordpress.com/2010/09/27/another-super-wrinkle-raising-typeerror/),  it's because the repeated calls to load_module act as a reload, the objects created using the old class no longer satisfy the isinstance test, so super will fail.
So remove super from code and call parent class instead.

### DIFF
--- a/src/collectors/ceph/ceph.py
+++ b/src/collectors/ceph/ceph.py
@@ -21,7 +21,7 @@ import glob
 import os
 import subprocess
 
-import diamond.collector
+from diamond.collector import Collector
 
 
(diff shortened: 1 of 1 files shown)


shortdudey123:
Since every collector uses the same super calls, I think you have run across a bigger issue.  Other collectors don't have this issue as far as I know.  If you still see this, open an issue.

PR #690, opened 2017-12-15
Changed custom version separator

acampoh:
I tried to generate yesterday a deb package with a custom version and i saw that debuild shows an error when trying to generate a package with a '-' in the version unless you use the **DEBUILD_TGZ_CHECK=no** config.


shortdudey123:
#714 added no tgzcheck.  Reopen if this is still a problem.

PR #698, opened 2018-02-28
Add option to disable staggering

DStape:


src/diamond/utils/scheduler.py
```diff
@@ -35,18 +36,24 @@ def collector_process(collector, metric_queue, log):
...
+    # Initially set stagger_offset to 0. If stagger_collection is enabled in
+    # the config, this value will soon be overridden.
+    stagger_offset = 0
+
     # Start the next execution at the next window plus some stagger delay to
     # avoid having all collectors running at the same time
     next_window = math.floor(time.time() / interval) * interval
-    stagger_offset = random.uniform(0, interval - 1)
+
+    if str_to_bool(self.config['stagger_collection']):
```
mattrobenolt, line 55: I'm guessing this needs to be `collector.config['stagger_collection']` instead of `self`.

But I have no capacity or ability to test this.

PR #707, opened 2018-05-04
Support for custom cloudwatch dimensions

jdonofrio728:
Provide support for custom dimensions for each rule (metric) by supporting a nested section called "dimensions".

Cloudwatch allows each metric to have up to 10 custom dimensions, that this PR will will to be configured.


jdonofrio728:
Does anyone monitor this project anymore?

PR #733, opened 2019-05-20
Correct link to documentation

mizhka:
Simple link correction for Ceph Perf counters.
Old link doesn't provide right page


mizhka:
Same as #727

PR #746, opened 2020-04-20
Add min_collection_time parameter to JolokiaCollector configuration

palnabarun:
Also, fixes some old ways of Python packaging.


palnabarun:
Closing this since the package seems to be unmaintained.

PR #760, opened 2022-01-02
Fix for issue introduced in #698

smartattack:
This patch should fix the error introduced with the merge of PR698:

Process MemoryCollector:
File "/usr/lib/python2.7/multiprocessing/process.py", line 267, in _bootstrap
self.run()
File "/usr/lib/python2.7/multiprocessing/process.py", line 114, in run
self._target(*self._args, **self._kwargs)
File "/home/petem/git/Diamond/src/diamond/utils/scheduler.py", line 55, in collector_process
if str_to_bool(self.config['stagger_collection']):
NameError: global name 'self' is not defined


smartattack:
Accidentally included a fix for issue introduced in PR759.  I meant to split these out as separate PRs, but I don't know how to undo it, offhand.

shortdudey123:
Thank you for the fix!

Issue #396, opened 2016-02-15
diamond is not py3k ready


sandrotosi:
python3 setup.py build  works fine but then running the result:

```
$ PYTHONPATH=build/lib/ python3 build/scripts-3.5/diamond
  File "build/scripts-3.5/diamond", line 112
    print "Diamond version %s" % (get_diamond_version())
                             ^
SyntaxError: invalid syntax
```

this is likely just the tip of the iceberg, and there is more work to do to make diamond really py3k ready

mattrobenolt:
I can say there's almost 0 chance this will happen and almost 0 value for the project. iirc, Diamond also supports down to 2.5, so I think it's almost impossible to support from 2.5 all the way through 3.5 since there are some fundamental syntax changes.

I could be wrong here, but I just don't see the value in Diamond needing to run in python3 either. It's not a library, so it should be self contained and shouldn't need to interop with other code.

sandrotosi:
python 2.5/2.6 are dead, so keeping their compatibility (and doing so preventing "progress") seems the wrong way for a python project.

other than then usual boilerplate of the advantage of py3k,  it is anyway the future of the language and a huge amount of work has been done to allow developers to write code that could be executed -without any change- both in 2.7 and 3.3+

There might be also people that would like to only run python3 on their machine (Arch switched /usr/bin/python to python3 long time ago), and given it's possible to extend with additional collectors, those might be easier written with a py3k-friendly syntax and/or you could gather more collaborations not closing that door in the code.

I wont use expressions like "0 value"/ no value, as it is discouraging and shortsighted.

thanks anyway for your attention

jaingaurav:
@sandrotosi: So I think support python 3 is important due to environments like Arch (and I think the latest Ubuntu as well). However transitioning to python3 is a significant undertaking beyond the core codebase. We need to ensure that collectors and their dependent libraries are python3 compatible, at least for the collectors that are regularly supported.

In the end, the real argument is time. Many of the diamond developers are strapped for time and we currently have some core priorities that need to be addressed first before taking on such an undertaking. If you are willing to spend the time to work on the changes, I'd be more than happy to help you in whatever ability I can.

sandrotosi:
[...]

PR #762, opened 2022-04-09
[wip] Python 3

erkie:
It felt kinda weird that there was absolutely nothing in this repo about how to run Diamond using Python 3. Like many others I recently upgraded to Ubuntu 20.04 which does not ship with a fully formed python 2.7 environment so I gave a stab at making it work with Python 3.

This is very much a WIP but I got it running fully for our setup.

How to install for you:

```
pip3 install distro
pip3 install git+https://github.com/feederco/Diamond.git@d8429765009fdf115c85aca09a5f2c1b570f8078
```

[Since `platform.distro()` was deprecated in 3.8](https://stackoverflow.com/questions/49554443/platform-linux-distribution-deprecated-what-are-the-alternatives) which `setup.py` relied on it now uses [distro](https://pypi.org/project/distro/), which is a pip module that needs to be installed before running pip install diamond.

Todo:
- Tests still not ported
- So far only tested in production with MySQL collector and hostedgraphite and archive handler
- This branch is all-out p3k, so we'd need to figure out how to interop these two languages, or tbh just make a fork, or keep python2 support in a branch, because python 2 seems pretty [ded](https://www.python.org/doc/sunset-python-2)

References #396


erkie:
Btw, since this is running fine in our environment I won't be working more on this. Any additional PR's and help is greatly appreciated.

shortdudey123:
Can you split out the diamond relative to full from /import statements like at the top of the handler files to a separate PR?  Those changes are non-breaking and I can merge it right away.  Also include the three lines changed in bin/diamond.

erkie:
@shortdudey123 I won't have time to do this anytime soon, sorry! Feel free to take over this branch or if any future readers see this please feel free to work off this branch 🙏

kt97679:
Thank you very much for the effort @erkie ! Based on your change I introduced more fixes to make diamond work in my environment: https://github.com/kt97679/Diamond/commit/8fee22d11f0928b2548c3f78dc3b3ae4b285f70a I would like to emphasize that I had to change config processing in the src/diamond/collector.py: https://github.com/kt97679/Diamond/commit/8fee22d11f0928b2548c3f78dc3b3ae4b285f70a#diff-65d20bb86255c9b2b68200bc4532b54b0c3fea580cf4ccaa0181c8259b2c2f6aL199-R208 I noticed that subsections using double square brackets are not parsed correctly so I switched from
```
[collectors]
...
[[default]]
```
to
```
[collectors]
...
[collectors.default]
```
Other changes are mostly automatic from the `2to3` script. In some places I had to add encode()/decode() logic.

PR #768, opened 2022-12-17
Minor netstat collector optimization for huge tables

aseev-xx:
- more pythonic, better and efficient way to split a string;
- do not load the entire table at once (memory savings for larger tables, over 150K+).

### DIFF
--- a/src/collectors/netstat/netstat.py
+++ b/src/collectors/netstat/netstat.py
@@ -46,25 +46,18 @@ class NetstatCollector(diamond.collector.Collector):
         Overrides the Collector.collect method
         """
 
-        content = self._load()
-        result = dict((self.STATE[num], 0) for num in self.STATE)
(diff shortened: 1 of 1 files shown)


src/collectors/netstat/netstat.py
```diff
@@ -46,26 +46,19 @@ def collect(self):
         Overrides the Collector.collect method
         """
 
-        content = self._load()
-        result = dict((self.STATE[num], 0) for num in self.STATE)
+        result = {self.STATE[num]: 0 for num in self.STATE}
```
aseev-xx, line 49: @shortdudey123 Hi!

I got feedback that in my patch, replacing dict() to {} led to incompatibility with python 2.6, which is a system one for CentOS6 and the collector breaks there.

If it seems important to you to maintain backward compatibility with old version (I could not find which versions are supported anywhere), then I will roll back this line. What do you think of it?

Thanks in advance for your reply.

PR #769, opened 2022-12-22
fix counter metric spike using statsD handler

cdong8812:
* Current implementation in statsD handler may work fine for general use case where the counter metric is sent from application directly to statsD (push), but in our prod env we have another stats application which pulls metrics from multiple applications and send to statsD, similar to how Promethus works (pull), and thus it will cause spike in the first data point every time when this stats application starts because the difference calculated automatically will be wrong if the actual application is already running for long time
* The fix is to adding an option for the caller to manually override the counter value using the value attribute but the internal update to the old value map will still use the raw value attribute


shortdudey123:
Can you modify only the needed if statement and not duplicate the entire code section including hasattr?

cdong8812:
Updated the PR

PR #770, opened 2023-02-22
fix bug: set user and group together:

ppa-arenadata:
ERROR: Failed to set UID/GID. global name 'user' is not defined


shortdudey123:
Please update this to use code similar to #654 since there is two sources for the user that can get used (see line 180).  Log message on line 224 should be updated too.

